Issue #26, opened 2015-02-15
toJava call on future converter causes java.lang.NoClassDefFoundError: scala/concurrent/java8/FuturesConvertersImpl$CF,


jasongoodwin:
Hey,

on v0.2.0 published to maven,
I've produced a barebones sbt project with only the scala-java8-compat library. 
You can recreate by running sbt test on this project:
https://github.com/jasongoodwin/future-converters-problem

There is a unit test that fails:

    @Test
    public void itShouldNotFail() {
        Future future = promise().future();
        final CompletionStage cs = toJava(future);
    }

It causes this exception

```
[error] Test pong.JavaAskExamplesTest.itShouldNotFail failed: java.lang.NoClassDefFoundError: scala/concurrent/java8/FuturesConvertersImpl$CF, took 0.062 sec
[error]     at scala.compat.java8.FutureConverters$.toJava(FutureConverters.scala:54)
[error]     at scala.compat.java8.FutureConverters.toJava(FutureConverters.scala)
[error]     at pong.JavaAskExamplesTest.itShouldNotFail(JavaAskExamplesTest.java:14)
[error]     ...
[error] Caused by: java.lang.ClassNotFoundException: scala.concurrent.java8.FuturesConvertersImpl$CF
[error]     at java.net.URLClassLoader$1.run(URLClassLoader.java:372)
[error]     at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
[error]     at java.security.AccessController.doPrivileged(Native Method)
[error]     at java.net.URLClassLoader.findClass(URLClassLoader.java:360)
[error]     at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
[error]     at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
[error]     ... 58 more
[error] Failed: Total 1, Failed 1, Errors 0, Passed 0
[error] Failed tests:
[error] 	pong.JavaAskExamplesTest
[error] (test:test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 32 s, completed 15-Feb-2015 10:43:38 AM
```

retronym:
Thanks for the report. Looks like the OSGi packaging in our build stripped those classes out of the final JAR. :cry: 

I'll cut a new release as soon as the fix is reviewed and merged.

jasongoodwin:
That's for publishing this so fast.

PR #27, opened 2015-02-16
Ensure scala.concurrent.java8._ is packaged

retronym:
The implementation of `FutureConverters` needs to access
some `private[concurrent]` members in the standard library.
To do so, it places some of its implementation under this package.

However, the OSGi SBT plugin seems to silently remove packages from the
final artifact that aren't either a publicly exported or a declared
private package.

Because of this, the 0.2.0 released JARs did not contain these
classes, and using FutureConverters resulted in a linkage error.

After this commit:

```
% sbt clean publishLocal

% jar tf target/scala-2.11/scala-java8-compat_2.11-0.2.0-SNAPSHOT.jar | grep -v 'scala/compat'
META-INF/MANIFEST.MF
scala-java8-compat.properties
scala/
scala/concurrent/
scala/concurrent/java8/
scala/concurrent/java8/FuturesConvertersImpl$.class
scala/concurrent/java8/FuturesConvertersImpl$CF$$anon$1.class
scala/concurrent/java8/FuturesConvertersImpl$CF.class
scala/concurrent/java8/FuturesConvertersImpl$P.class
scala/concurrent/java8/FuturesConvertersImpl.class
```

As compared with:

```
% curl --silent https://oss.sonatype.org/content/repositories/releases/org/scala-lang/modules/scala-java8-compat_2.11/0.2.0/scala-java8-compat_2.11-0.2.0.jar | tar tf - | grep -v 'scala/compat'
META-INF/MANIFEST.MF
scala-java8-compat.properties
scala/
```

Review by @rkuhn

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,6 +19,8 @@ scalaModuleOsgiSettings
 
 OsgiKeys.exportPackage := Seq(s"scala.compat.java8.*;version=${version.value}")
 
+OsgiKeys.privatePackage := List("scala.concurrent.java8.*")
+
 libraryDependencies += "junit" % "junit" % "4.11" % "test"
 
 libraryDependencies += "com.novocode" % "junit-interface" % "0.10" % "test"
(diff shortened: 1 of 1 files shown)


bantonsson:
LGTM

retronym:
Fixes #26

PR #30, opened 2015-04-28
Optionals compat

Ichoran:
Converters between Java classes `Optional`, `OptionalDouble`, `OptionalInt`, and `OptionalLong`; and Scala `Option`.  Tests and docs included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,29 @@ class Test {
  - [API](src/main/java/scala/compat/java8/FutureConverters.java)
  - [Test Cases](src/test/java/scala/compat/java8/FutureConvertersTest.java)
 
+## Converters between `scala.Option` and `java.util` classes `Optional`, `OptionalDouble`, `OptionalInt`, and `OptionalLong`.
+
+A set of extension methods to enable explicit conversion between [Scala Option](http://www.scala-lang.org/files/archive/api/2.11.6/#scala.Option) and the Java 8
+optional types, [Optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html),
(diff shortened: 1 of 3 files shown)


retronym:
This project cross builds against 2.10 and 2.11, so we have to use the lowest common denominator:

```
[error] /home/travis/build/scala/scala-java8-compat/src/main/scala/scala/compat/java8/OptionConverters.scala:101: value class needs to have a publicly accessible val parameter
[error]   implicit class RichOptionalLong(private val o: OptionalLong) extends AnyVal {
[error]                                               ^
```

Either publicize the parameter (obfuscating the name to `_self`), or manually expand the implicit class to a regular class and an implicit def.

retronym:
Otherwise LGTM. Could you please squash these into one commit before we merge (I'd rather have the tests and doc land in the same commit as the code.)

Ichoran:
@retronym - Manually expanding the implicit class isn't any use, as it is 2.10s insistence on public vals that pollutes the namespace.  I typically call everything `underlying` in that case, but I could use `_self` if that's a more widely adopted standard.

Ichoran:
@retronym - Squashed, made all value class names public as "underlying", and added toJava/toScala methods for use from Java.  But maybe I should have another namespace and make those be "toJavaOptional", "toJavaOptionalDouble", etc., in more traditional Java style?  (So from Java you can just import them all and not worry about the Scala clutter?)

retronym:
I'm happy to focus on the Scala API side in this case, I'll merge as is. Let's come back in a second pass to see if there are opportunities to improve consistency across our various converters. (We are still pre-1.0 in the library.)

PR #31, opened 2015-04-29
Conversions from Scala and Java Iterators to PrimitiveIterator

Ichoran:
Specifically, to OfDouble, OfInt, and OfLong.

Docs and tests included.

No explicit Java support, but new scala.compat.java8.wrappers.IteratorPrimitiveIntWrapper(i) is not _that_ hard.

### DIFF
--- a/src/main/java/scala/compat/java8/wrappers/IteratorPrimitiveDoubleWrapper.java
+++ b/src/main/java/scala/compat/java8/wrappers/IteratorPrimitiveDoubleWrapper.java
@@ -1,0 +1,13 @@
+package scala.compat.java8.wrappers;
+
+public class IteratorPrimitiveDoubleWrapper implements java.util.PrimitiveIterator.OfDouble {
+  private java.util.Iterator<Double> it;
+  public IteratorPrimitiveDoubleWrapper(java.util.Iterator<Double> it) {
+    this.it = it;
+  }
+  public boolean hasNext() { return it.hasNext(); }
+  public double nextDouble() { return it.next().doubleValue(); }
+  public void forEachRemaining(java.util.function.Consumer<? super Double> c) {
+    it.forEachRemaining(c);
+  }
+}
--- a/src/main/scala/scala/compat/java8/PrimitiveIteratorConversions.scala
+++ b/src/main/scala/scala/compat/java8/PrimitiveIteratorConversions.scala
@@ -1,0 +1,108 @@
+package scala.compat.java8
+
+import language.implicitConversions
+import java.util.{ Iterator => JIterator, PrimitiveIterator }
+
+/** This class enables conversion from `scala.Iterator` to the set of
+  * `java.util.PrimitiveIterator` classes.
+  *
+  * Scala's `Iterator` is generic, as is its `java.util` counterpart.  However,
+  * `java.util.PrimitiveIterator` offers three manually-specialized variants of
+  * `Iterator`: `OfDouble`, `OfInt`, and `OfLong`.  This class provides
+  * `.asPrimitive` extension methods for Scala and Java iterators to present
+  * the generic versions as the specialized version.
+  *
+  * Example usage:
+  *
+  * {{{
+  * import scala.compat.java8.PrimitiveIteratorConverters._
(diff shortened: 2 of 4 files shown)


src/main/java/scala/compat/java8/wrappers/IteratorPrimitiveDoubleWrapper.java
```diff
@@ -0,0 +1,13 @@
+package scala.compat.java8.wrappers;
+
+public class IteratorPrimitiveDoubleWrapper implements java.util.PrimitiveIterator.OfDouble {
```
retronym: Should we make these final?

src/main/java/scala/compat/java8/wrappers/IteratorPrimitiveDoubleWrapper.java
```diff
@@ -0,0 +1,13 @@
+package scala.compat.java8.wrappers;
```
retronym: Perhaps `package scala.compat.java8.collection`?

src/main/scala/scala/compat/java8/PrimitiveIteratorConversions.scala
```diff
@@ -0,0 +1,108 @@
...
+  *
+  * Example usage:
+  *
+  * {{{
+  * import scala.compat.java8.PrimitiveIteratorConverters._
+  * val it = Iterator(1.0, 2.0, math.Pi)
+  * val jpid = it.asPrimitive   // PrimitiveIterator.OfDouble
+  * }}}
+  */
+object PrimitiveIteratorConverters {
```
retronym: Is this intentionally in a different package than the wrapper classes themselves?

Ichoran:
@retronym - After I wrote this, I was thinking that it would be nicest to get all the Java 8 collections compatibility stuff with import scala.collection.java8._ and not have to worry about what is in and what is out.  (And get scala.compat.java8._ for everything else.)  So I probably ought to move everything into a package class for scala.collection.java8.  (For streams compatibility we'll need something in scala.collection because a lot of stuff is marked private[collection] which we need access to.)

src/main/java/scala/compat/java8/wrappers/IteratorPrimitiveDoubleWrapper.java
Ichoran: Maybe `scala.collection.java8.primitiveCompat` or somesuch, if we move it into the scala.collection namespace?

src/main/java/scala/compat/java8/wrappers/IteratorPrimitiveDoubleWrapper.java
Ichoran: I left them non-final since that seemed to be idiomatic Java.  (The JIT compiler is really good at detecting monomorphic access.)  But the implementation is trivial; it wouldn't hurt to make them final either.  Might be a good signal that this just a particular implementation that we happen to be using.

src/main/java/scala/compat/java8/wrappers/IteratorPrimitiveDoubleWrapper.java
retronym: I was hoping to keep all API from this module under `scala.compat`, rather than trying to augment standard library packages.

The basis of this decision was originally to avoid shipping "split packages". Thinking about it some more, this might be less of a problem if we just add a package in a stdlib package, rather than directly add classes.

Here's the original discussion: https://groups.google.com/a/typesafe.com/forum/#!search/scala-java8-fun/eng/iUyQxzcsGfs/3jvl6cfTYhoJ

retronym:
Maybe we can structure the implicits into traits:

```scala
trait OptionConverters { implicit def ... }
object ObjectConverters

trait IteratorConverters { implicit def ... }
object IteratorConverters

object AllConverters extends OptionConverters with IteratorConverters
```

retronym:
I'm not in favour of adding implicits to the package object though.

retronym:
I'm going to merge this, and we can spend some time to figure out the organization details as a followup.

retronym:
> Docs and tests included.

I think the tests got lost along the way.

PR #35, opened 2015-05-05
Update build.sbt

retronym:


retronym:
Forgot to do this last time, hence the two version bump.

PR #37, opened 2015-05-13
Add a generic deserializer for Java/Scala 2.12 lambdas

retronym:
Java support serialization of lambdas by using the serialization
proxy pattern. Deserialization of a lambda uses `LambdaMetafactory`
to create a new anonymous subclass.

More details of the scheme are documented:

  https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/SerializedLambda.html

From those docs:

> SerializedLambda has a readResolve method that looks for a
> (possibly private) static method called $deserializeLambda$
> in the capturing class, invokes that with itself as the first
> argument, and returns the result. Lambda classes implementing
> $deserializeLambda$ are responsible for validating that the
> properties of the SerializedLambda are consistent with a lambda
> actually captured by that class.

The Java compiler generates code in `$deserializeLambda$` that
switches on the implementation method name and signature to locate
an invokedynamic instruction generated for the particular lambda
expression. Then, the `SerializedLambda` is further unpacked,
validating that this implementation method still represents the
same functional interface as it did when it was serialized.
(The source may have been recompiled in the interim.)

In Java, serializable lambda expressions are the exception rather than
the rule. In Scala, however, the serializability of `FunctionN` means
that we would end up generating a large amount of code to support
deserialization.

Instead, we are pursuing an alternative approach in which the
`$deserializeLambda$` method is a simple forwarder to the generic
deserializer added here.

This is capable of deserializing lambdas created by the Java compiler,
although this is not its intended use case. The enclosed tests use
Java lambdas.

This generic deserializer also works by calling `LambdaMetafactory`,
but it does so explicitly, rather than implicitly during linkage
of the `invokedynamic` instruction.

We have to mimic the caching property of `invokedynamic` instruction
to ensure we reuse the classes when constructing. The cache here
uses weak references to keys and values to avoid retention of `Class`
or `ClassLoader` instances.

If the name or signature of the implementation method has changed,
we fail during deserialization with an `IllegalArgumentError.`

However, we do not fail fast in a few cases that Java would, as we
cannot reflect on the "current" functional interface supported by
this implementation method. We just instantiate using the "previous"
functional interface class/method.

This might:

1. fail inside `LambdaMetafactory` if the new implementation
method is not compatible with the old functional interface.

2. pass through `LambdaMetafactory` by chance, but fail
when instantiating the class in other cases. For example:

```
% tail sandbox/test{1,2}.scala
==> sandbox/test1.scala <==
class C {
  def test: (String => String) = {
    val s: String = ""
    (t) => s + t
  }
}

==> sandbox/test2.scala <==
class C {
  def test: (String, String) => String = {
    (s, t) => s + t
  }
}
% (for i in 1 2; do scalac -Ydelambdafy:method -Xprint:delambdafy sandbox/test$i.scala 2>&1 ; done) | grep 'def $anon'
    final <static> <artifact> private[this] def $anonfun$1(t: String, s$1: String): String = s$1.+(t);
    final <static> <artifact> private[this] def $anonfun$1(s: String, t: String): String = s.+(t);
```

3. Silently create an instance of the old functional interface.
For example, imagine switching from `FuncInterface1` to
`FuncInterface2` where these were identical other than the name.

I don't believe that these are showstoppers.


retronym:
Review by @lrytz

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
```diff
@@ -0,0 +1,112 @@
...
+   *
+   * This cache is weak in keys and values to avoid retention of the enclosing class (and its classloader)
+   * of deserialized lambdas.
+   *
+   * Assumptions:
+   *  - No additional marker interfaces are required beyond `{java.io,scala.}Serializable`. These are
+   *    not stored in `SerializedLambda`, so we can't reconstitute them.
+   *  - No additional bridge methods are passed to `altMetafactory`. Again, these are not stored.
+   *
+   * Note: The Java compiler
```
lrytz: incomplete :information_source:

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
```diff
@@ -0,0 +1,112 @@
...
+    val loader = lookup.lookupClass().getClassLoader
+    val implClass = loader.loadClass(slashDot(serialized.getImplClass))
+
+    def makeCallSite: CallSite = {
+      import serialized._
+      def parseDescriptor(s: String) =
+        MethodType.fromMethodDescriptorString(s, loader)
+
+      val funcInterfacesSignature = parseDescriptor(getFunctionalInterfaceMethodSignature)
+      val methodType: MethodType = funcInterfacesSignature
```
lrytz: seems unused

src/test/java/scala/compat/java8/runtime/LambdaDeserializerTest.java
```diff
@@ -0,0 +1,108 @@
...
+    B apply(A a);
+}
+
+class LambdaHost {
+    public F1<Boolean, String> lambdaBackedByPrivateImplMethod() {
+        int local = 42;
+        return (b) -> Arrays.asList(local, b ? "true" : "false", LambdaHost.this).toString();
+    }
+
+    public F1<Boolean, String> lambdaBackedByStaticImplMethod() {
```
lrytz: this looks the same as the above `lambdaBackedByPrivateImplMethod`, in both cases the lambda body seems to be non-static

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
```diff
@@ -0,0 +1,112 @@
...
+      val instantiated = parseDescriptor(getInstantiatedMethodType)
+      val implMethodSig = parseDescriptor(getImplMethodSignature)
+
+      val from = implMethodSig.parameterCount() - funcInterfacesSignature.parameterCount()
+      val to = implMethodSig.parameterCount()
+      val functionalInterfaceClass = loader.loadClass(slashDot(getFunctionalInterfaceClass))
+      var invokedType: MethodType =
+        implMethodSig.dropParameterTypes(from, to)
+          .changeReturnType(functionalInterfaceClass)
```
lrytz: why is this `changeReturnType` needed?

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
```diff
@@ -0,0 +1,112 @@
...
+          throw new IllegalArgumentException("Illegal lambda deserialization", e)
+      }
+      val FLAG_SERIALIZABLE = 1
+      val FLAG_MARKERS = 2
+      val flags: Int = FLAG_SERIALIZABLE | FLAG_MARKERS
+      val markerInterface: Class[_] = if (functionalInterfaceClass.getName.startsWith("scala.Function"))
+        loader.loadClass("scala.Serializable")
+      else
+        loader.loadClass("java.io.Serializable")
+
```
lrytz: So `FLAG_SERIALIZABLE` seems to already [add the `java.io.Serializable` interface](https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/LambdaMetafactory.html#altMetafactory-java.lang.invoke.MethodHandles.Lookup-java.lang.String-java.lang.invoke.MethodType-java.lang.Object...-).

We should check what happens if we also add `scala.Serializable` as a marker interface - do we get both? Is that OK?

lrytz:
The code and the approach LGTM, really nice.

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
retronym: I'll comment this part. We are looking at the implementation method:

Given:

```
val x: FunctionalInterface<C, D, E> = (c, d) => { println(capture1, capture2); new E }
```

Which is compiled to be backed by an implementation method like:

```
def impl(capture1: C1, capture2: C2, param1: C, param2: D): LambdaResultType
```

And deriving the effective type of the lambda factory.

```
(C1, C2)FunctionalInterface
```

Or without erasure: `(C1, C2) => FunctionalInterface[C, D, E]`

I like to think of the impl method as being written as:

```
def impl(capture1: C1, capture2: C2)(param1: C, param2: D): LambdaResultType
```

And then of LMF as eta expansion.

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
retronym: I will try removing the explicit addition of "java.io.Serializable" and test that is is added. But looking that the bytecode that Javac emits and the implementation of `LambdaMetafactory`, I think that documentation is misleading, and the marker interfaces must be explicitly added. Conversely, adding a the marker interface that extends Serializable without setting the flag is not sufficient, that will consider it to be "accidentally serializable", and generate a `writeObject` that throws.

Although `scala.Serializable` was intended as a JVM-independent way to mark a Scala class a serializable, in practice people can also write methods that depend an instance of that trait, so we definitely need to keep it here.  It would be cleaner if it were a straight type alias.

retronym: The docs are correct, I was looking at the wrong spot in the implementation. I'm about to push a commit that explains that it is redundant but safe to add `java.io.Serializable`.

src/test/java/scala/compat/java8/runtime/LambdaDeserializerTest.java
retronym: I've fixed this and confirmed we get test coverage of the different paths. I've also added some variations of Java lambdas that were not working before, e.g. `String::new` or `Object::toString` or `Clonable::clone`.

retronym:
Ready for re-review, @lrytz

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
lrytz: got it, thanks

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
```diff
@@ -0,0 +1,135 @@
+package scala.compat.java8.runtime
+
+import java.lang.invoke._
+import java.lang.ref.WeakReference
+
+/**
+ * This class is only intended to be called by synthetic `$deserializeLambda$` method that the Scala 2.12
+ * compiler will add to classes hosting lambdas.
+ *
+ * It is intended to be consumed directly.
```
lrytz: Not sure what message this phrase is conveying :) - or does it miss a "not"?

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
```diff
@@ -0,0 +1,135 @@
...
+ */
+object LambdaDeserializer {
+  private final case class CacheKey(implClass: Class[_], implMethodName: String, implMethodSignature: String)
+  private val cache = new java.util.WeakHashMap[CacheKey, WeakReference[CallSite]]()
```
lrytz: I was thinking a if we should bound the size of this cache. I guess the answer is no, but it'd still be interested to hear what you think about it.

If I count correctly, there will be one entry per lambda-instantiation-site for which one of its created lambda objects is ever deserialized.

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
```diff
@@ -0,0 +1,135 @@
...
+   * is cheap, it amounts to a reflective call to the constructor of the previously created class.
+   * However, deserialization of the same lambda expression is not guaranteed to use the same class,
+   * concurrent deserialization of the same lambda expression may spin up more than one class.
+   *
+   * This cache is weak in keys and values to avoid retention of the enclosing class (and its classloader)
+   * of deserialized lambdas.
+   *
+   * Assumptions:
+   *  - No additional marker interfaces are required beyond `{java.io,scala.}Serializable`. These are
+   *    not stored in `SerializedLambda`, so we can't reconstitute them.
```
lrytz: One thing I still don't quite understand: why do we need to add the `Serializable` marker interface if the functional interface extends `Serializable` anyway? The created lambda object has a class type that implements the functional interface type, so it would have the right `Serializable` already.

retronym: The `scala.FunctionN` traits don't extends `Serializable`, but the subclasses synthesized for anonymous functions do.

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
```diff
@@ -0,0 +1,135 @@
...
+   *  - No additional bridge methods are passed to `altMetafactory`. Again, these are not stored.
```
lrytz: Just confirming: for normal lambda creation (not deserialization), we use the ordinary `metafactory` (not `altMetafactory`) at the indy-callsite, which never creates additional bridge methods in the lambda object?

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
retronym: I don't think it should be bounded. But your comment prompted me to reconsider this cache, and I've found a problem with it: while the weak references to the keys prevents the classic problem of strong references from the value back to the weak key undermining the attempted weakness of the cache, it also makes the key eligible for garbage collection sooner than we would like.

https://github.com/retronym/scala-java8-compat/compare/topic/lambda-deserialize...retronym:topic/lambda-deserialize-weak

So rather than `WeakHashMap[CacheKey, WeakReference[CallSite]]`, we need something like `WeakHashMap[CacheKey, WeakReference[Class] => CallSite]`.

retronym: Or maybe we don't need weak values. I'm going to try to make a test that loads a lambda hosting class in a nested classloader. I'll then create a lambda hosted by that class, serialize and deserialize it, drop the reference to the classloader, churn GC, and see if the cache entry clears. My worry is that the strong reference to the lambda class will prevent GC of its containing classloader.

retronym: Here an instance of the sort of issue I'm trying to avoid: https://github.com/FasterXML/jackson-databind/issues/489

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
retronym: The companion PR in scala/scala switched to using `altMetafactory` unconditionally.

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
lrytz: Ah, that's new in the scala/scala PR, currently it just uses `metafactory` and `JFunctionN`, right?

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
retronym: Correct :smile:

retronym:
Yep, that's right. Note that JFUnctionN doesn't implement Serializable, either.

retronym:
s/faithfully unknown/faithfully deserialize unknown/ above.

retronym:
See also: 

http://permalink.gmane.org/gmane.comp.java.openjdk.core-libs.devel/4895
https://bugs.openjdk.java.net/browse/JDK-6493635

retronym:
One thing we could consider is synthesizing a static cache per-class. We could then reason about the object lifetimes by analogy to the reflection caches for structural type invocations.

src/main/java/scala/compat/java8/runtime/LambdaDeserializer.scala
retronym: I've decided to go with the approach of hosting a per-class deserialization cache.

lrytz:
LGTM!

PR #39, opened 2015-06-02
Added function converter capability.

Ichoran:
Based on initial reflection work from Jason Zaugg.

If you have something from java.util.functions, you should be able to

```
import scala.compat.java8.FunctionConverters._
```

and then use `.asScala` to convert to the corresponding Scala function.

If you have a Scala function, use `.asJava` to convert to the best-matching item from java.util.functions.

The code is created with a code generator that needs to run in Scala 2.11+, so build.sbt was reworked to be multi-module (the sub-module is the code generator).


retronym:
@Ichoran The build failure is due to the lack of quasiquotes in Scala 2.10. I'm going to drop 2.10 support at this point in a separate PR, so we can disregard those.

retronym:
I'd like to the approach to implicit priorities documented and tested.

Here's one example that is ambiguous:

```
scala> ((x: Any) => "").asJava
<console>:45: error: type mismatch;
 found   : Any => String
 required: ?{def asJava: ?}
Note that implicit conversions are not applicable because they are ambiguous:
 both method wrap_asJava_IntFunction in trait Priority1FunctionConverters of type [R](scalaFn: Int => R)scala.compat.java8.WrappedAsJava[java.util.function.IntFunction[R]]
 and method wrap_asJava_LongFunction in trait Priority1FunctionConverters of type [R](scalaFn: Long => R)scala.compat.java8.WrappedAsJava[java.util.function.LongFunction[R]]
 are possible conversion functions from Any => String to ?{def asJava: ?}
              ((x: Any) => "").asJava
                        ^
<console>:45: error: value asJava is not a member of Any => String
              ((x: Any) => "").asJava
                               ^
```

Ichoran:
@retronym - Hmm, I hadn't anticipated that (but should have).  It's breaking the most-specific-type part of the computation.  But then because of the inversion of specificity due to covariance of function arguments, I wonder whether I need the prioritization at all?  I'll play around with it this weekend.  I agree I need tests and docs.  I don't recall why I pushed this before writing them--maybe I wanted feedback?  (Thanks!)

Ichoran:
Hmm, I hadn't anticipated that (but should have).  Contravariance of function arguments makes things trickier than usual.  I might need to add some type identity computations to get it to work.  I'll work on it this weekend.  (Also--I know I need tests and docs.  I think I might have pushed this because I wanted feedback?--I don't recall.  Anyway, thanks!)

Ichoran:
@retronym - I did a fairly extensive reworking of the PR, and added tests.  I think this is now a decent candidate for inclusion (after appropriate review and modification).

retronym:
Continued in #48

PR #41, opened 2015-06-05
Make bridge interfaces implement serializable

retronym:
I'm trying to use this library to help with some Java8 code written on top of the Spark API.  I'm using a spark library that takes scala Function1s.  Unfortunately, I cannot use the scala-java8-compat library because the bridge interface does not extend Serializable and for a java8 lambda to be serializable the target type has to be serializable: http://docs.oracle.com/javase/tutorial/java/javaOO/lambdaexpressions.html#serialization


retronym:
[Converted issue to a pull request]

punya:
This would be really useful to me, for the same reason.

PR #46, opened 2015-07-28
Add Scala 2.12.0-M2 as a cross build

retronym:


retronym:
Review by @rkuhn or @viktorklang

retronym:
Marking as "on-hold" as this would prevent #50

viktorklang:
@retronym The reason for final-ing DefaultPromise is to make sure that the optimizer can do its job, avoiding tripping into bimorphic or megamorphic callsites. I think the extra allocation and indirection in your proposal is OK given that, but also to prevent others from extending it. Wdyt?

retronym:
@viktorklang Can you see a way to implement #50 if `DefaultPromise` if final? The only solution I see is to copy/paste its implementation.

retronym:
We could mark a selection of methods like `isCompleted` and `value` as final `DefaultPromise` to keep them bimorphic (`KeptPromise` also implements them).

viktorklang:
@retronym Perhaps an option would be to make all of the methods on DefaultPromise final?

retronym:
Yeah, I'm fine with that.

viktorklang:
@retronym We also need to audit the methods to make sure that nothing will break (is there something that assumes no inheritance?) as well as make sure that methods that may have harmful effects if called from inheriters(?) will be either hidden or appropriately documented.

retronym:
Yep, changing modifiers is subtle. Could you review your addition of the finality to confirm it was done solely for non functional reasons?

retronym:
I don't think we need to lock it down to much as it is the whole class already access restricted. If someone plays a trick to subvert that (as we do here) we assume they are responsible for the consequences

viktorklang:
I can review the addition at the earliest next week :/

2m:
Released latest genjavadoc (which is `0.9`) for scala 2.12.0-M2. Can you update to 0.9 and keep the build clean of scala version switch statements? :) I tried it locally and there seems to be no problems when using `0.9`.

retronym:
ping @viktorklang for assessment on the impact of unfinalizing `DefaultPromise`.

retronym:
@2m Thanks for that. I've updated this PR to use 0.9 of genjavadoc in both 2.11 and 2.12.

viktorklang:
@retronym I only had time for a cursory overview, it seems as unfinalizing it will be fine. May be in order to final-ize all of its methods tho.

Issue #29, opened 2015-02-21
Can't use thenCompose with FutureConvertersImpl - doThenCompose calls toCompletableFuture


jasongoodwin:
Hey I found an issue with this.
The FutureConvertersImpl throws an exception when toCompletableFuture is called:

    override def toCompletableFuture(): CompletableFuture[T] ={
      throw new UnsupportedOperationException("this CompletionStage represents a read-only Scala Future")
    }

Which is fine. However, if you do something like this:

    @Test
    public void testToJavaThenComposeWithToJavaThenAccept() throws InterruptedException,
            ExecutionException, TimeoutException {
        final Promise<String> p1 = promise();
        final CompletableFuture future = new CompletableFuture();

        final CompletionStage<String> second =
                CompletableFuture.supplyAsync(() -> "Hello").
                        thenCompose(x -> toJava(p1.future()));

        second.handle((x, t) -> {
            if(t!=null){
                t.printStackTrace();
                future.completeExceptionally(t);
            }else{
                future.complete(x);
            }
            return null;
        });

        p1.success("Hello");
        assertEquals("Hello", future.get(1000, MILLISECONDS));
    }

thenCompose calls doThenCompose:

    public <U> CompletableFuture<U> thenCompose
        (Function<? super T, ? extends CompletionStage<U>> fn) {
        return doThenCompose(fn, null);
    }

And if there aren't any exceptions, then it calls toCompletableFuture which causes failure. 
Removing the exception causes everything to complete just fine.
Is there another way this can be handled other than throwing an exception?

        if (ex == null) {
                if (e != null) {
                    if (dst == null)
                        dst = new CompletableFuture<U>();
                    execAsync(e, new AsyncCompose<T,U>(t, fn, dst));
                }
                else {
                    try {
                        CompletionStage<U> cs = fn.apply(t);
                        if (cs == null ||
                            (dst = cs.toCompletableFuture()) == null)
                            ex = new NullPointerException();
                    } catch (Throwable rex) {
                        ex = rex;
                    }
                }
            }

    private <U> CompletableFuture<U> doThenCompose
        (Function<? super T, ? extends CompletionStage<U>> fn,
         Executor e) {
        if (fn == null) throw new NullPointerException();
        CompletableFuture<U> dst = null;
        ThenCompose<T,U> d = null;
        Object r;
        if ((r = result) == null) {
            dst = new CompletableFuture<U>();
            CompletionNode p = new CompletionNode
[...]

PR #49, opened 2015-07-28
Support blocking with: `scalaFuture.toJava.toCompletableFuture`

retronym:
Thought I'd get the ball rolling with a PR.

Review by @jroper / @viktorklang / @rkuhn (as available)

### DIFF
--- a/src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
+++ b/src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
@@ -72,9 +72,11 @@ object FuturesConvertersImpl {
      * WARNING: completing the result of this method will not complete the underlying
      *          Scala Future or Promise (ie, the one that that was passed to `toJava`.)
      */
-    override def toCompletableFuture(): CompletableFuture[T] = {
-      this
-    }
+    override def toCompletableFuture(): CompletableFuture[T] = this
+
+    override def obtrudeValue(value: T): Unit = throw new UnsupportedOperationException("obtrudeValue may not be used on the result of toJava(scalaFuture)")
+
+    override def obtrudeException(ex: Throwable): Unit = throw new UnsupportedOperationException("obtrudeException may not be used on the result of toJava(scalaFuture)")
 
     override def get(): T = scala.concurrent.blocking(super.get())
 
--- a/src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
+++ b/src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
@@ -66,8 +66,15 @@ object FuturesConvertersImpl {
       cf
     }
 
-    override def toCompletableFuture(): CompletableFuture[T] =
-      throw new UnsupportedOperationException("this CompletionStage represents a read-only Scala Future")
+    /**
+     * @inheritdoc
+     *
+     * WARNING: completing the result of this method will not complete the underlying
+     *          Scala Future or Promise.
+     */
+    override def toCompletableFuture(): CompletableFuture[T] = {
+      this // TODO or maybe `thenApply(JF.identity())`
+    }
 
     override def toString: String = super[CompletableFuture].toString
   }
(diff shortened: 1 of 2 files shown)


src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
```diff
@@ -66,8 +66,15 @@ object FuturesConvertersImpl {
       cf
     }
 
-    override def toCompletableFuture(): CompletableFuture[T] =
-      throw new UnsupportedOperationException("this CompletionStage represents a read-only Scala Future")
+    /**
+     * @inheritdoc
+     *
+     * WARNING: completing the result of this method will not complete the underlying
```
viktorklang: There is no underlying Scala Future or Promise for this class? (CF)

viktorklang: What happens if the CF is completed by a Scala Future prior to `toCompletableFuture.complete(x)` gets called?

retronym: > There is no underlying Scala Future or Promise for this class? (CF)

The underlying future is the one passed to `toJava`:

```
  def toJava[T](f: Future[T]): CompletionStage[T] = {
    val cf = new CF[T]
    implicit val ec = InternalCallbackExecutor
    f onComplete cf
    cf
  }
```

In my other PR, that also becomes a field of `CF` so that we can efficiently unwrap it.

retronym: > What happens if the `CF` is completed by a Scala Future prior to `toCompletableFuture.complete(x)` gets called?

I've added a test case to show what happens, could you please review it?

src/test/java/scala/compat/java8/FutureConvertersTest.java
```diff
@@ -3,321 +3,377 @@
...
-      }
-      return CompletableFuture.completedFuture(x);
-    });
-    p.success("Hello");
-    latch.countDown();
-    assertEquals("Hello", second.toCompletableFuture().get());
-  }
+        second.handle((x, t) -> {
+            if (t != null) {
+                t.printStackTrace();
```
viktorklang: Is this intentional?

src/test/java/scala/compat/java8/FutureConvertersTest.java
```diff
@@ -3,321 +3,377 @@
...
+        CompletableFuture<String> cf = cs.toCompletableFuture();
+        assertEquals("notyet", cf.getNow("notyet"));
+        cf.complete("done");
+        assertEquals("done", cf.get());
+        assertFalse(sf.isCompleted());
+        assertFalse(p.isCompleted());
+    }
 
+    @Test
+    public void testToJavaToCompletableFutureJavaCompleteCallledAfterScalaComplete() throws ExecutionException, InterruptedException {
```
viktorklang: Is there a test for the inverse?

src/test/java/scala/compat/java8/FutureConvertersTest.java
retronym: I copied the test verbatim from #29. I have just pushed another commit that notes its origin and simplifies it a little.

src/test/java/scala/compat/java8/FutureConvertersTest.java
retronym: Just added one.

src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
```diff
@@ -66,8 +66,15 @@ object FuturesConvertersImpl {
...
+     *          Scala Future or Promise (ie, the one that that was passed to `toJava`.)
+     */
+    override def toCompletableFuture(): CompletableFuture[T] = {
+      this // TODO or maybe `thenApply(JF.identity())`
```
viktorklang: this TODO isn't really helpful, is it?

src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
viktorklang: That test looks great.

What happens if `obtrudeX` is called?

viktorklang:
One thought, shouldn't `get()` + `get(timeout)` use scala.concurrent.blocking?

src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
retronym: Removed.

retronym:
> One thought, shouldn't get() + get(timeout) use scala.concurrent.blocking?

Sounds legit. Does my last commit put this in the right spot?

retronym:
@viktorklang Please see my latest commit that addresses the `obtrude`.

PR #50, opened 2015-07-28
More efficient round trip conversions for Futures

### LINKED ISSUE #44: Wrapping future
I'm wondering why `FutureConverters.toJava` creates a new `CompletableFuture`, rather than simply wrapping the `Future` with an implementation of `CompletionStage` that delegates to `Future`?  Doing this would make conversion a little more efficient, and would mean conversions back and forth could be optimised so a conversion from a previously converted `CompletionStage` to `Future` would simply unwrap the `Future`.

The latter especially will be useful in Play, since we want to allow both Java and Scala code to intercept requests, and mixing these interceptors will potentially mean lots of conversions back and forth between the types.  If it's just a simple wrapping/unwrapping, then we don't need to worry about the cost associated with it, otherwise we may end up with long chains of Futures and CompletionStages that add nothing to the actual value except overhead.

retronym:
A straw man proposal, for discussion by @jroper @rkuhn and
@viktorklang.

See the second commit for details, as I've based this PR on
another in-flight PR.

Fixes #44

### DIFF
--- a/src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
+++ b/src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
@@ -5,8 +5,9 @@ package scala.concurrent.java8
 
 // Located in this package to access private[concurrent] members
 
-import scala.concurrent.{ Future, Promise, ExecutionContext, ExecutionContextExecutorService, ExecutionContextExecutor, impl }
+import scala.concurrent.{ Future, ExecutionContext }
 import java.util.concurrent._
+import scala.concurrent.impl.Promise.DefaultPromise
 import scala.util.{ Try, Success, Failure }
 import java.util.function.{ BiConsumer, Function ⇒ JF, Consumer, BiFunction }
 
@@ -14,7 +15,7 @@ import java.util.function.{ BiConsumer, Function ⇒ JF, Consumer, BiFunction }
 object FuturesConvertersImpl {
   def InternalCallbackExecutor = Future.InternalCallbackExecutor
 
-  class CF[T] extends CompletableFuture[T] with (Try[T] => Unit) {
+  class CF[T](val wrapped: Future[T]) extends CompletableFuture[T] with (Try[T] => Unit) {
     override def apply(t: Try[T]): Unit = t match {
       case Success(v) ⇒ complete(v)
       case Failure(e) ⇒ completeExceptionally(e)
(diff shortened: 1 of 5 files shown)


src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
```diff
@@ -6,7 +6,7 @@ package scala.concurrent.java8
 // Located in this package to access private[concurrent] members
 
 import scala.concurrent.{ Future, Promise, ExecutionContext, ExecutionContextExecutorService, ExecutionContextExecutor, impl }
-import java.util.concurrent.{ CompletionStage, Executor, ExecutorService, CompletableFuture }
+import java.util.concurrent._
```
viktorklang: Aren't you shadowing scala.concurrent.Future now?

src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
```diff
@@ -66,13 +66,20 @@ object FuturesConvertersImpl {
       cf
     }
 
-    override def toCompletableFuture(): CompletableFuture[T] =
-      throw new UnsupportedOperationException("this CompletionStage represents a read-only Scala Future")
+    /**
+     * @inheritdoc
+     *
+     * WARNING: completing the result of this method will not complete the underlying
```
viktorklang: Ah, now I see it, so you are using 2 words for the same Future, use `underlying` or `wrapped` consistently please :)

src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
```diff
@@ -66,13 +66,20 @@ object FuturesConvertersImpl {
...
+     */
+    override def toCompletableFuture(): CompletableFuture[T] = {
+      this // TODO or maybe `thenApply(JF.identity())`
+    }
 
     override def toString: String = super[CompletableFuture].toString
   }
 
-  class P[T] extends impl.Promise.DefaultPromise[T] with BiConsumer[T, Throwable] {
+  class P[T](val wrapped: CompletionStage[T]) extends impl.Promise.DefaultPromise[T] with BiConsumer[T, Throwable] {
```
viktorklang: What if `wrapped` is a `CF`?

retronym: The only place that creates instance of `P` would not be reached with `cs : CF`:

```scala
   def toScala[T](cs: CompletionStage[T]): Future[T] = {
    cs match {
      case cf: CF[T] => cf.wrapped
      case _ =>
        val p = new P[T](cs)
        cs whenComplete p
        p.future
    }
  }
```

Do you disagree? What problems do you envisiage if wrapped *was* a `CF`?

src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
retronym: Nope, wildcards have a lower precendence:

```
scala> {import scala.concurrent.Future; import java.util.concurrent._; 0 : Future[_]}
<console>:8: error: type mismatch;
 found   : Int(0)
 required: scala.concurrent.Future[_]
              {import scala.concurrent.Future; import java.util.concurrent._; 0 : Future[_]}
                                                                              ^
```

src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
retronym: will do

jroper:
This looks good to me.

PR #60, opened 2015-12-22
Enable cross build for 2.12.0-M3

retronym:


SethTisue:
if you published this, can you add it to https://github.com/scala/make-release-notes/blob/2.12.x/projects-2.12.md ?

PR #61, opened 2016-01-08
Compatibility for Java 8 Streams through Steppers, a Spliterator/Iterator hybrid

Ichoran:
This provides basic support for getting Java 8 Streams from Scala collections by defining a new iterator-like type, Stepper, that is simultaneously a Spliterator (for Stream compatibility) and an Iterator (since Scala collections can all be iterated anyway).  It also provides a large Array-like accumulator type, Accumulator (and specialized variants thereof) that can be used to very rapidly collect the results of a Java 8 Stream (comparable to Array, but with unbounded size) and then operate on them as a Stream again.

Documentation has been extended to cover the use cases on both of these.

A partial set of benchmarks test a variety of common use cases to make sure performance is acceptable.  (It is; in general Steppers used as Iterators are as fast or faster than Iterator, with a notable exception for both kinds of Queue at the moment.)


README.md
```diff
@@ -117,6 +117,105 @@ class Test {
...
+#### Scala Usage Example
+
+```scala
+import scala.compat.java8.StreamConverters._
+
+object Test {
+  val m = collection.immutable.HashMap("fish" -> 2, "bird" -> 4)
+  val s = m.parValueStream.sum          // 6, potientially computed in parallel
+  val t = m.seqKeyStream.toScala[List]  // List("fish", "bird")
+  val a = t.accumulate                  // Accumulator[(String, Int)]
```
lrytz: should probably be `m.accumulate` instead of `t.accumulate`

README.md
```diff
@@ -117,6 +117,105 @@ class Test {
...
+
+  val n = a.splitter.fold(0)(_ + _._1.length) +
```
lrytz: is `splitter` a typo?

Ichoran: Yeah, stepper.  Thanks for spotting that!  At some point I called it "splitter".  I need to get in the habit of rerunning example code as a last step.

Ichoran:
@lrytz - I've updated and completed the docs.  @adriaanm - What should I do next to get this adequately reviewed?  It's a big project, and unfortunately I have few cases where I need to use this myself.  (Maybe I should invent some.)

metasim:
I'd like this capability, but am not qualified to review the pull request. Perhaps merge and create an "RC" release?

adriaanm:
I'll see who's best placed to spend some quality time with this. We're going to push out M4 a bit more because of some unforeseen complications with the trait encoding work, so we'll have plenty of time to review this PR in time for M4 (should be out before end of March).

README.md
```diff
@@ -117,6 +117,134 @@ class Test {
 ```
 
 
-## Future work
-  - Converters for `java.util.stream`
-  - [`Spliterator`](https://docs.oracle.com/javase/8/docs/api/java/util/Spliterator.html)s for Scala collections
+## Converters from Scala collections to Java 8 Streams
+
+Scala collections gain `seqStream` and `parStream` as extension methods that produce a Java 8 Stream
```
szeiger: I'm wondering if `seq` and `par` in Scala collections should be renamed to `sequential` and `parallel`. AFAICT they see little use in practice, `seq` is confusing (because we also have a `Seq` type) q

Ichoran: I wonder the same thing, but I don't think it matters much to this API.  We can have the names be as above or `sequentialStream` and `parallelStream` or whatever regardless.

src/main/scala/scala/compat/java8/collectionImpl/Stepper.scala
```diff
@@ -0,0 +1,545 @@
+package scala.compat.java8.collectionImpl
+
+import java.util.Spliterator
+
+/** A Stepper is a specialized collection that can step through its
```
szeiger: What's the value in exposing this as a collection type? Wouldn't it suffice to keep it an implementation detail and only expose spliterators and streams?

README.md
```diff
@@ -117,6 +117,134 @@ class Test {
...
+running sequentially or in parallel, respectively.  These are automatically specialized to a primitive
+type if possible.  For instance, `List(1,2).seqStream` produces an `IntStream`.  Maps additionally have
+`seqKeyStream`, `seqValueStream`, `parKeyStream`, and `parValueStream` methods.
+
+Scala collections also gain `accumulate` and `stepper` methods that produce utility collections that
```
szeiger: When would you use these methods? It looks like `accumulate` always copies the collection sequentially, which does not seem useful to me as a separate feature. For sequential streams, I'd expect that I could just use the original collection, and for parallel streams I'd want the implementation to decide whether copying is really necessary or the original Scala parallel collection (or even a sequential collection like ArrayBuffer) can be used efficiently. Building an accumulator sequentially could still be done with `to[Accumulator]`, which would probably set the right expectations for performance characteristics.

src/main/scala/scala/compat/java8/collectionImpl/Stepper.scala
Ichoran: The Spliterator interface is low-performance compared to the next/hasNext interface of Iterators.  Since Stepper provides the latter, it can be valuable to have access to it.

szeiger: How does that differ from using the existing `.iterator` method to get an Iterator for a collection?

Ichoran: Normal iterators can't be split effectively.  It's a primitive collection from which you can perform parallel operations.  Spliterator is another idea for how to do it, but it doesn't perform as well due to difficulties in fully inlining the side-effecting closure.

README.md
szeiger: I can see why you'd want to use it internally, I'm just concerned about exposing new concepts like `Accumulator` and `Stepper` in the API at this point. If there's no compelling use-case I'd just leave them internal and see if a need for more API arises.

Ichoran: Adriaan asked me to write a framework for exposing more collections internals so it was easier to do things like write parallelizing code.  It seemed that having a Stepper already there would be a benefit.  But I agree, it could probably be hidden as an implementation detail (at least with `private[scala.compat]`).

I'm not so sure it makes sense to hide Accumulator.  If you want to safely collect intermediate results of unknown size, it is the thing to use.  There are no equivalents.

szeiger: Right, for `Accumulator` I would only argue for removal of the `accumulator` convenience methods on collections. The type needs to be exposed as part of the API.

README.md
```diff
@@ -117,6 +117,134 @@ class Test {
...
+
+For sequential operations, Scala's `iterator` almost always equals or exceeds the performance of a Java 8 stream.  Thus,
+one should favor `iterator` (and its richer set of operations) over `seqStream` for general use.  However, long
+chains of processing of primitive types can sometimes benefit from the manually specialized methods in `DoubleStream`,
+`IntStream`, and `LongStream`.
+
+Note that although `iterator` typically has superior performance in a sequential context, the advantage is modest
+(usually less than 50% higher throughput for `iterator`).
+
+For parallel operations, `parStream` and even `seqStream.parallel` meets or exceeds the performance of Scala parallel
```
szeiger: Are there use cases for having `seqStream.parallel` be different from `parStream`? I would expect a sequential stream that was constructed directly from a collection to return that collection's native parallel stream when you call `.parallel`.

README.md
```diff
@@ -117,6 +117,134 @@ class Test {
...
+def mapToSortedString[A](xs: Vector[A], f: A => String, sep: String) =
+  xs.parStream.
+    map[String](a => f(a)).sorted.  // Explicit lambda creates a SAM wrapper for f
+    toArray.mkString(sep)
+```
+
+#### Java Usage Example
+
+To convert a Scala collection to a Java 8 Stream from within Java, it usually
+suffices to call `ScalaStreaming.from(xs)` on your collection `xs`.  If `xs` is
```
szeiger: How about calling this `ScalaStream.of(xs)` or `ScalaStreamSupport.stream(xs)` to stay closer to the Java APIs `Stream.of` and `StreamSupport.stream`?

Ichoran: That's a good idea.

README.md
Ichoran: If you are sure that you're going to support parallel operations, you might want to do different work up front.  I don't remember if I actually implemented that, but note that the steppers do _not_ see the `.parallel` call; all they know is that at some point `trySplit` gets called on the `Spliterator` interface.

README.md
Ichoran: Fair enough.  I'm not sure inconvenience helps more than simplification, but I don't have a strong feeling either way.

src/main/java/scala/compat/java8/ScalaStreaming.java
```diff
@@ -0,0 +1,905 @@
+package scala.compat.java8;
+
+import scala.compat.java8.converterImpl.*;
+import scala.compat.java8.collectionImpl.*;
+import java.util.stream.*;
+import scala.compat.java8.runtime.CollectionInternals;
+
+public class ScalaStreaming {
```
szeiger: This could use some comments here or on this individual methods about immutability, or otherwise the binding policy of the generated streams.

src/main/java/scala/compat/java8/runtime/CollectionInternals.java
```diff
@@ -0,0 +1,23 @@
+package scala.compat.java8.runtime;
+
+import scala.collection.immutable.*;  // Don't rely on this!  Refer to everything explicitly!
```
szeiger: Why import it at all?

src/main/scala/scala/compat/java8/collectionImpl/Stepper.scala
```diff
@@ -0,0 +1,545 @@
...
+        }
+        sub
+      }
+    }
+    override def tryAdvance(c: java.util.function.LongConsumer) = useCache(c) || sp.tryAdvance(c)
+  }
+  
+  /** Creates a `Stepper` over a generic `Spliterator`. */
+  def ofSpliterator[A](sp: Spliterator[A]): AnyStepper[A] = sp match {
+    case as: AnyStepper[A] => as
```
szeiger: What about specialized Steppers? Would it make sense to have a separate implementation that wraps them as AnyStepper instead of falling back to Spliterator?

src/main/scala/scala/compat/java8/StreamConverters.scala
```diff
@@ -0,0 +1,339 @@
...
+
+/** `StreamConverters` provides extension methods and other functionality to
+  * ease interoperability of Scala collections with `java.util.stream` classes.
+  * 
+  * Scala collections gain extension methods `seqStream` and
+  * `parStream` that allow them to be used as the source of a `Stream`.
+  * Some collections either intrinsically cannot be paralellized, or
+  * could be but an efficient implementation is missing.  It this case,
+  * only `seqStream` is provided.  If a collection cannot be stepped over
+  * at all (e.g. `Traversable`), then it gains neither method.
```
szeiger: Why wouldn't `TraversableOnce` be sufficient? Anthing that can produce an `Iterator` could also produce a `Stepper`, right?

src/main/scala/scala/compat/java8/collectionImpl/Accumulator.scala
```diff
@@ -0,0 +1,344 @@
...
+        hIndex += 1
+      }
+      index = that.index
+      current = that.current
+    }
+    totalSize += that.totalSize
+    that.clear
+  }
+  
+  override def clear(): Unit = {
```
szeiger: What's the use case for `clear` (and the fact that `drain` calls it)? Wouldn't you only use `drain` in a merger where the original RHS gets discarded anyway?

src/main/scala/scala/compat/java8/collectionImpl/Accumulator.scala
```diff
@@ -0,0 +1,344 @@
...
+  final def spliterator: java.util.Spliterator[A] = stepper.spliterator
+  
+  /** Produces a sequential Java 8 Stream over the elements of this `Accumulator`*/
+  final def seqStream: java.util.stream.Stream[A] = java.util.stream.StreamSupport.stream(spliterator, false)
+  
+  /** Produces a parallel Java 8 Stream over the elements of this `Accumulator`*/
+  final def parStream: java.util.stream.Stream[A] = java.util.stream.StreamSupport.stream(spliterator, true)
+  
+  /** Copies the elements in this `Accumulator` into an `Array` */
+  final def toArray(implicit tag: reflect.ClassTag[A]) = {
```
szeiger: Why not use `System.arrayCopy` for this?

src/main/scala/scala/compat/java8/collectionImpl/Accumulator.scala
```diff
@@ -0,0 +1,344 @@
...
+      }
+      h -= 1
+    }
+    ans
+  }
+  
+  /** Copies the elements in this `Accumulator` to a specified collection.
+    * Usage example: `acc.to[Vector]`
+    */
+  final def to[Coll[_]](implicit cbf: collection.generic.CanBuildFrom[Nothing, A, Coll[A]]): Coll[A] = {
```
szeiger: Could this be moved up to `AccumulatorLike`? Scala specialization should be able to produce the correct specialized versions for the subclasses automatically.

src/main/scala/scala/compat/java8/collectionImpl/IntAccumulator.scala
```diff
@@ -0,0 +1,346 @@
...
+  }
+  
+  private def loadMore(): Unit = {
+    h += 1
+    if (h < acc.hIndex) { a = acc.history(h); n = acc.cumulative(h) - acc.cumulative(h-1) }
+    else { a = acc.current; n = acc.index }
+    i = 0
+  }
+  
+  def characteristics = ORDERED | SIZED | SUBSIZED
```
szeiger: Add `NONNULL` (in all primitive AccumulatorSteppers)?

src/main/scala/scala/compat/java8/collectionImpl/Stepper.scala
```diff
@@ -0,0 +1,545 @@
...
+  def tryStep(f: A => Unit): Boolean
+
+  /** Attempt to split this `Stepper` in half, with the new (returned) copy taking the first half
+    * of the collection, and this one advancing to cover the second half.  If subdivision is not
+    * possible or not advisable, `substep()` will return `null`.
+    */
+  def substep(): CC
+
+  /** Returns the precise underlying type of this `Stepper`. */
+  def typedPrecisely: CC
```
szeiger: Why use a separate method here instead of a self-type on `StepperLike`?

src/main/scala/scala/compat/java8/collectionImpl/Stepper.scala
```diff
@@ -0,0 +1,545 @@
...
+    */
+  def iterator: Iterator[A] = new scala.collection.AbstractIterator[A] {
+    def hasNext = self.hasStep
+    def next = self.nextStep
+  }
+
+  /** Returns a Scala collection of the type requested. */
+  def to[Coll[_]](implicit cbf: collection.generic.CanBuildFrom[Nothing, A, Coll[A]]): Coll[A] = {
+    val b = cbf()
+    while (hasStep) b += nextStep
```
szeiger: Could this be faster with `foreach`, which is overridden in some `StepperLike` subclasses in a more efficient way?

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/** Classes or objects implementing this trait create generic steppers suitable for sequential or parallel use. */
+trait MakesAnyStepper[A] extends Any {
+  /** Generates a fresh stepper over `A`s that can be efficiently subdivided */
+  def stepper: AnyStepper[A] with EfficientSubstep
+}
+
+/** Classes or objects implementing this trait create generic steppers for map keys suitable for sequential or parallel use. */
+trait MakesAnyKeyStepper[A] extends Any {
```
szeiger: Can you abstract over `EfficientSubstep` in a useful way so that it makes sense to have a separate set of traits that mandate this marker trait? I can't think of any good use cases for this. If a method can make good use of either subdivisible or sequential steppers, it can make the distinction at runtime. If sequential processing would degrade performance, what are you going to do if you get a stepper without `EfficientSubstep`? It's probably still your best option.

szeiger: Thinking further: Even if you want it prevent inefficient steppers statically, you could ask for a function that returns the right stepper instead of taking an object that implements one of the `Make...` traits. That would also allow abstracting over key / value / sequence steppers.

src/main/scala/scala/compat/java8/converterImpl/StepsLikeGapped.scala
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
+
+/** Abstracts the process of stepping through an incompletely filled array of `AnyRefs`
+  * and interpreting the contents as the elements of a collection.
+  */
+private[java8] abstract class StepsLikeGapped[A, STA >: Null <: StepsLikeGapped[A, _]](_underlying: Array[AnyRef], _i0: Int, _iN: Int)
+  extends AbstractStepsLikeGapped[AnyStepper[A], STA](_underlying, _i0, _iN)
+  with AnyStepper[A]
+  with EfficientSubstep
```
szeiger: Move `with EfficientSubstep` up to `AbstractStepsLikeGapped`? The same pattern occurs for a few other classes.

src/main/scala/scala/compat/java8/converterImpl/StepsString.scala
```diff
@@ -0,0 +1,44 @@
...
+    else null
+  }
+}
+
+/////////////////////////
+// Value class adapter //
+/////////////////////////
+
+final class RichStringCanStep(private val underlying: String) extends AnyVal with MakesIntStepper {
+  @inline def stepper: IntStepper with EfficientSubstep = new StepperStringCodePoint(underlying, 0, underlying.length)
```
szeiger: I would expect a method on `String` that is simply called `stepper` to return a Stepper over chars, not codepoints. Also, calling `.iterator` on a `String` now goes through the Stepper, so you get an Iterator over codepoints instead of chars.

src/test/scala/scala/compat/java8/StepConvertersTest.scala
```diff
@@ -0,0 +1,572 @@
...
+    assert(correctSpec(x))
+    assert(!isAcc(x))
+    assert(!isLin(x))
+  }
+
+  def Tell[X](x: => X)(implicit correctSpec: SpecCheck) {
+    println(x.getClass.getName + " -> " + isAcc(x))
+    assert(x.isInstanceOf[Stepper[_]])
+    assert(correctSpec(x))
+  }
```
szeiger: Some reuse could make the seemingly arbitrary methods easier to understand, e.g.:

```
  def Okay[X](x: => X)(implicit correctSpec: SpecCheck) {
    Fine(x)
    assert(isLin(x))
  }

  def good[X](x: => X)(implicit correctSpec: SpecCheck) {
    Fine(x)
    assert(!isLin(x))
  }
```

src/test/scala/scala/compat/java8/StepConvertersTest.scala
```diff
@@ -0,0 +1,572 @@
...
+    Okay( cm.Stack[String]("salmon").stepper )
+    IFFY( cm.Traversable[String]("salmon").accumulate.stepper )
+    Okay( cm.TreeSet[String]("salmon").stepper )
+    Okay( cm.UnrolledBuffer[String]("salmon").stepper )
+    Okay( cm.WeakHashMap[String, String]("fish" -> "salmon").stepper )
+    Okay( cm.WeakHashMap[String, String]("fish" -> "salmon").keyStepper )
+    Okay( cm.WeakHashMap[String, String]("fish" -> "salmon").valueStepper )
+    good( (Array("salmon"): cm.WrappedArray[String]).stepper )
+
+    // Java 6 converters section
```
szeiger: What's supposed to go into these empty Java 6 converter sections?

src/main/scala/scala/compat/java8/StreamConverters.scala
```diff
@@ -0,0 +1,339 @@
...
+    def accumulatePrimitive[AA](implicit psa: PrimitiveStreamAccumulator[S, AA]) = psa.streamAccumulate(stream)
+  }
+}
+
+/** `StreamConverters` provides extension methods and other functionality to
+  * ease interoperability of Scala collections with `java.util.stream` classes.
+  * 
+  * Scala collections gain extension methods `seqStream` and
+  * `parStream` that allow them to be used as the source of a `Stream`.
+  * Some collections either intrinsically cannot be paralellized, or
```
Ichoran: Typoed `parallelized`!

src/main/scala/scala/compat/java8/StreamConverters.scala
Ichoran: This is to be more explicit about where work is happening.  If you want to copy the entire object, that's your job (preferably into an Accumulator); silently copying a potentially huge collection does not seem like a good idea to me.

src/main/scala/scala/compat/java8/collectionImpl/Accumulator.scala
Ichoran: It's to match the capabilities of `Buffer` (without being constrained by its inheritance hierarchy).  It is probably inessential.  (Or at least can be worked around by nulling out a whole Accumulator instead of its contents.)

src/main/scala/scala/compat/java8/collectionImpl/Accumulator.scala
Ichoran: The array might contain primitives.  Doing it this way is safe, if inefficient.  Maybe should special-case the AnyRef case.

src/test/scala/scala/compat/java8/StepConvertersTest.scala
Ichoran: I think I was supposed to write wrappers for the pre-Java-8 collections wrappers.  And I don't think I did.

src/test/scala/scala/compat/java8/StepConvertersTest.scala
Ichoran: I'm not sure the lack of reuse is the problem.  The naming is horrible but was fast to type and easy to spot at a glance (when you know what to look for) as I switched one or two at a time from not-good to good as I implemented more of the features.  I should probably search-and-replace it all to more descriptive names.

src/main/scala/scala/compat/java8/converterImpl/StepsString.scala
Ichoran: Ack.  Yeah.  This was an early proof of concept.  Maybe I should remove it entirely.  If not, it should be something like codePointStepper.  What do you think?

src/main/scala/scala/compat/java8/converterImpl/StepsLikeGapped.scala
Ichoran: That should be safe.

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
Ichoran: So, the way this was supposed to work was that things that had `EfficientSubstep` could be run in a parallel stream just by asking for it, whereas if you don't have it you have to explicitly ask for a sequential stream and call the parallel method on it (or bite the bullet and turn it into an accumulator before you begin).  It was really for the stream layer, not for the steppers themselves.  And I had it as a marker trait so it can decorate both AnyStepper, which is inherently an iterator that implements tryAdvance and so on, and the spliterator-based steppers.  But I'm not sure I took advantage of that, so a redesign might be the way to go.

I am pretty sure that the Makes stuff was part of the type computation to hook up the right collections with the right streams, but otherwise the function idea is better.  I should check to make sure.

src/main/scala/scala/compat/java8/collectionImpl/Stepper.scala
Ichoran: That's a good point.  Avoiding the mandatory caching as well as the boxing is probably a win.

src/main/scala/scala/compat/java8/collectionImpl/Stepper.scala
Ichoran: I don't remember, I'm afraid.  It was a bit tricky to get the manual specialization working and integrated properly with the automatic specialization, and it's possible that this was an artifact of my explorations there that ended up being necessary at some point (I am sure it was) but isn't any more.  Or maybe it was for Java interop?  I guess I could experiment with dropping it.

src/main/scala/scala/compat/java8/collectionImpl/Stepper.scala
Ichoran: It's a tradeoff.  `foreach` has an extra closure, which makes the optimization harder in the case where it's just doing this.  Since in my implementations it almost always does just this, I went with this one.  I don't think I benchmarked extensively enough to know what is better and worse by how much, though.

src/main/java/scala/compat/java8/ScalaStreaming.java
Ichoran: That's a good idea.

src/main/java/scala/compat/java8/runtime/CollectionInternals.java
Ichoran: That is a very good question.  I'll remove it and see if anything breaks.

src/main/scala/scala/compat/java8/collectionImpl/Accumulator.scala
Ichoran: I am pretty sure I checked and that in practice it didn't actually work as well.  But I can check again and say what I find.

src/main/scala/scala/compat/java8/collectionImpl/IntAccumulator.scala
Ichoran: Good point!

Ichoran:
@szeiger - I've commented on everything you said, but haven't changed any source files yet.  I'll do that this weekend.  Some of the larger issues I may not have time to get to (e.g. typedPrecisely vs. trait self-type), but I'll let you know.

README.md
szeiger: Good point. You could override `Stream.parallel` to make the stepper know but that would only work for the initial stream, so it's probably not worth the extra complexity.

src/main/scala/scala/compat/java8/StreamConverters.scala
szeiger: I agree. I thought `TraversableOnce` had an `iterator` method but it doesn't, so it shouldn't have a Stepper, either.

src/main/scala/scala/compat/java8/collectionImpl/Accumulator.scala
szeiger: You mean copying from an `Accumulator[Int]` which contains buffers of type `Array[java.lang.Integer]` to an `Array[Int]`? Could this case be prevented by using the right buffer type in the first place?

src/main/scala/scala/compat/java8/converterImpl/StepsString.scala
szeiger: Do we need a `Stepper` at all? Are you supposed to use those directly in Scala code, not just as a building block for Java Streams? If so, `codePointStepper` would be a good name, otherwise I'd remove it entirely because `CharSequence` already has a `codePoints` method that produces an `IntStream`.

szeiger:
I think the focus for now should be on API-related issues that would easily break source compatibility if we were to change them later, e.g. availability of accumulator methods or the codepoint vs char Stepper on String. Anything else like removing typedPrecisely, adding doc comments, or moving specialized methods up into a parent class for simplification could also be done later. We don't really need any of that for the first release or for this PR, which is already rather unwieldy.

szeiger:
Here are the changes to remove `typedPrecisely`: https://github.com/szeiger/scala-java8-compat/commit/8ca53f0a210d1e196986f436dbf23c06bf966e57

I looked at the implementation of `trySplit` in `abstract class DoubleStepperTest extends DoubleStepper`. With `typedPrecisely`:

```
   public java.util.Spliterator trySplit();
    Code:
       0: aload_0
       1: invokevirtual #356                // Method trySplit:()Ljava/util/Spliterator$OfDouble;
       4: areturn

  public java.util.Spliterator$OfPrimitive trySplit();
    Code:
       0: aload_0
       1: invokevirtual #356                // Method trySplit:()Ljava/util/Spliterator$OfDouble;
       4: areturn

   public java.util.Spliterator$OfDouble trySplit();
    Code:
       0: aload_0
       1: invokestatic  #62                 // Method scala/compat/java8/collectionImpl/DoubleStepper$class.trySplit:(Lscala/compat/java8/collectionImpl/DoubleStepper;)Ljava/util/Spliterator$OfDouble;
       4: areturn
```

Without:

```
  public java.util.Spliterator trySplit();
    Code:
       0: aload_0
       1: invokevirtual #354                // Method trySplit:()Lscala/compat/java8/collectionImpl/DoubleStepper;
       4: areturn

  public java.util.Spliterator$OfPrimitive trySplit();
    Code:
       0: aload_0
       1: invokevirtual #354                // Method trySplit:()Lscala/compat/java8/collectionImpl/DoubleStepper;
       4: areturn

  public java.util.Spliterator$OfDouble trySplit();
    Code:
       0: aload_0
       1: invokevirtual #354                // Method trySplit:()Lscala/compat/java8/collectionImpl/DoubleStepper;
       4: areturn

  public scala.compat.java8.collectionImpl.DoubleStepper trySplit();
    Code:
       0: aload_0
       1: invokestatic  #62                 // Method scala/compat/java8/collectionImpl/DoubleStepper$class.trySplit:(Lscala/compat/java8/collectionImpl/DoubleStepper;)Lscala/compat/java8/collectionImpl/DoubleStepper;
       4: areturn
```

So there's an additional method with a `DoubleStepper` return type but both versions have the same number of indirections and should work equally well with specialization.

src/main/scala/scala/compat/java8/StreamConverters.scala
```diff
@@ -0,0 +1,339 @@
...
+  * `Array` also gains `seqStream` and `parStream` methods, and calling those
+  * on `Array[Double]`, `Array[Int]`, or `Array[Long]` will produce the
+  * corresponding primitive stream.
+  *
+  * Streams gain `accumulate` and `toScala[_]` methods, which collect the stream
+  * into a custom high-performance `scala.collection.mutable.java8.Accumulator`,
+  * which is not part of the standard collections hierarchy, or into a named
+  * Scala collection, respectively.
+  *
+  * Generic streams also gain an `unboxed` method that will convert to the
```
szeiger: Should we allow such streams in the first place? AFAIK the Java APIs provide a nice (albeit manual) separation of primitive and generic types, so any such stream would have to come from Scala. How about we restrict the `seqStream` and `parStream` methods to element types `<: AnyRef` and remove the `unboxed` and `boxed` methods? If people really want to use boxed streams for primitive values, they'd have to cast the collection to an AnyRef element type.

szeiger: Ah, I see. The `boxed` method actually comes from Java, so it makes sense to have a corresponding `unbox`. I'd still restrict the types for StreamConverters though.

Ichoran: Almost all the Java benchmarks showing Java 8 Streams wiping the floor with Scala are using primitive streams.  I think they should be as deeply supported as possible.  In particular, giving the relative difficulty of optimizing tryAdvance-based code, just throwing in an extra unboxing step is _not_ just like throwing in an extra method call to do it.  It actually adds quite a bit of overhead in my tests.  (I forget exactly how much.)

szeiger: Right, which is why I would prevent the creation of boxed streams for primitive types in the first place. If you still want to do it, you have to cast to the boxed type.

Ichoran: But what if you want a stream of `Char`?  There is no primitive type that corresponds.  Do we want to make this intentionally awkward?  (What do we even do?)

I already have designed the priority hierarchy such that if you have a collection of Double you can't conveniently get a Stream instead of a DoubleStream out of it.  (I hope.  That was my goal.)

szeiger: > But what if you want a stream of Char? There is no primitive type that corresponds. Do we want to make this intentionally awkward? (What do we even do?)

The supported primitive stream types can support all other primitive types through widening conversions, so I'd say use `IntStream` for `Char`. To make this feasible, we'd also need a way to perform a narrowing conversion when collecting into an unboxed primitive collection (like `Array[Char]`).

> I already have designed the priority hierarchy such that if you have a collection of Double you can't conveniently get a Stream instead of a DoubleStream out of it. (I hope. That was my goal.)

The loophole is abstracting over types. The current design allows you to get a Stream for a collection of `T <: Any` and since it's implemented through an extension method with static dispatch, all primitive types are boxed. Restricting the supported types to `T <: AnyRef` would make that illegal, so you're forced to special-case primitive types (or cast if you're willing to accept the boxing overhead).

szeiger: Case in point: `CharSequence.chars()` returns an `IntStream`

szeiger:
Some optimizations for the extension methods: https://github.com/szeiger/scala-java8-compat/commit/e6edfa898ff743b468e876f5caa4374e5f59c0f6

src/main/scala/scala/compat/java8/converterImpl/StepsString.scala
szeiger: Optional char and codePoint steppers for String: https://github.com/szeiger/scala-java8-compat/commit/0835e860badaa95ad76dcf4d1347b1f456706b5e

src/main/scala/scala/compat/java8/converterImpl/StepsLikeGapped.scala
szeiger: Implementation: https://github.com/szeiger/scala-java8-compat/commit/4c4855f0b8d8b7c20efa2310e0383c1f5a26ccf1

src/main/scala/scala/compat/java8/converterImpl/StepsString.scala
retronym: Could we not build upon (or just let people directly use `CharSequence#{chars, codePoints}`, rather than implementing our own steppers here? If we do need to implement our own, would it not make sense to generalize to `CharSequence` rather than `String`?

retronym: Oh sorry, just saw that I've duplicated @szeiger's earlier comment here.

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
szeiger: Here's a first pass at simpification: https://github.com/szeiger/scala-java8-compat/commit/daf8bd692e5b2fcd1210ca88ff67446d17de0f94

The separate traits for key / value / generic steppers are still necessary but it now abstracts over the stepper types. The way these typeclasses are used for producing streams, I don't think we can get rid of them easily. If we moved the conversion from steppers to streams down into the typeclasses, it should be possible to remove the special cases for primitive stream converters.

src/main/scala/scala/compat/java8/collectionImpl/IntAccumulator.scala
szeiger: Implementation: https://github.com/szeiger/scala-java8-compat/commit/498994faba6b79ca576e78b35764ffac86f38852

README.md
szeiger: Implementation: https://github.com/szeiger/scala-java8-compat/commit/f5988a77244e902995322a50bf0f2f081ffdd59c

src/main/java/scala/compat/java8/runtime/CollectionInternals.java
szeiger: Seems to work: https://github.com/szeiger/scala-java8-compat/commit/f190438f698deed09a26013a87acd5617dc54362

src/main/java/scala/compat/java8/ScalaStreaming.java
szeiger: Here's my take on it: https://github.com/szeiger/scala-java8-compat/commit/cd37db5bb1ba2041fa01c7233452201bc4f4cfee

src/main/scala/scala/compat/java8/collectionImpl/Accumulator.scala
szeiger: The only place that calls `drain` is `Accumulator.merger` for use by `Stream.collect`. It is not clear to me from the documentation of `collect` whether implementations are supposed to drop references quickly. I would assume that they generally do though, otherwise all *immutable* accumulation would always retain references to old objects. In particular, the example in the documentation is `stringStream.collect(ArrayList::new, ArrayList::add, ArrayList::addAll)`. This would suffer from the same problem. OTOH, not clearing the RHS Accumulator would leave it in an invalid state. It's probably not worth complicating the API with such a "dangerous" method for the little performance gain that can be expected.

src/main/scala/scala/compat/java8/collectionImpl/Stepper.scala
szeiger: Implementation: https://github.com/szeiger/scala-java8-compat/commit/e627f8373d342f81879a905e0752256d1fc2d520

szeiger:
Since this looks mostly good, with only a few details left to investigate, I'm going to merge it and then create a cleaned-up PR with some changes that I expect to be uncontroversial. This is an experimental module and the upcoming release is a milestone, so we can still do change to the API.

PR #65, opened 2016-03-31
Improvements to Java 8 Stream support

szeiger:
This is a collection of changes from the discussion on #61 which I expect to be straight-forward. Compared to my earlier branch, I removed the commit that added the `T <: AnyRef` restriction to `StreamConverters`. This needs a better design to take widening conversions into account.

Review by @Ichoran


Ichoran:
LGTM.  I didn't check every single site of modification, but the major cleanups (separating parStream and seqStream and flattening the chain of low priority traits, and changing the parameterization of MakesStepper) look really good.

PR #66, opened 2016-04-01
Automatic widening of numeric types for primitive Streams

szeiger:
The widening conversions follow the usual pattern. Float becomes
Double, Byte/Char/Short become Int. Boolean and Unit are not widened.

For Arrays, where the actual data is unboxed, widening is implemented
through custom Stepper implementations, so you can get e.g. a native
IntStepper for an Array[Byte]. The StreamConverters hierarchy is
modified to prefer these widened primitive steppers, so that, for
example, the `seqStream` extension method on an Array[Byte] returns
an IntStream.

Other collection types with a boxed representation also get the same
widening conversions (so you can still benefit from the better
performance of unboxed streams) but there are no implicitly available
widening steppers. Only the Stream extension methods fall back to
applying a widening stepper in order to produce a widening primitive
stream.

Widening narrow numeric types for streams is consistent with the Streams
API in the JDK. For example, `CharSequence.chars` returns an `IntStream`
instead of a `Stream[Character]`.

### DIFF
--- a/src/test/scala/scala/compat/java8/StepConvertersTest.scala
+++ b/src/test/scala/scala/compat/java8/StepConvertersTest.scala
@@ -438,6 +438,23 @@ class StepConvertersTest {
   }
 
   @Test
+  def shortWidening() {
+    implicit val spec = SpecCheck(_.isInstanceOf[IntStepper])
+
+    good( Array[Short](654321.toShort).stepper )
+
+    //TODO: None of these currently work because there are no native Stepper implementations. This does not only
+    // affect widening conversions though. While you can get, for example, an IntStepper for a WrappedArray[Int],
+    // all values have to go through a boxing/unboxing step!
+
+    //good( ci.NumericRange(123456.toShort, 123458.toShort, 1.toShort).stepper )
+    //good( ((Array[Short](654321.toShort): cm.WrappedArray[Short]): cm.ArrayLike[Short, cm.WrappedArray[Short]]).stepper )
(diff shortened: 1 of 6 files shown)


src/test/scala/scala/compat/java8/StepConvertersTest.scala
```diff
@@ -438,6 +438,23 @@ class StepConvertersTest {
...
 
   @Test
+  def shortWidening() {
+    implicit val spec = SpecCheck(_.isInstanceOf[IntStepper])
+
+    good( Array[Short](654321.toShort).stepper )
+
+    //TODO: None of these currently work because there are no native Stepper implementations. This does not only
+    // affect widening conversions though. While you can get, for example, an IntStepper for a WrappedArray[Int],
+    // all values have to go through a boxing/unboxing step!
```
szeiger: @Ichoran, I think this part brings us back to the role of Steppers in the API and whether they should be exposed to the user. If so, we may want to provide widening conversions at the Stepper level for convenience, otherwise we could probably remove many of the existing primitive Stepper implementations which are based on boxed collections, and apply an unboxing Stepper similar to the new widening stepper adapters.

szeiger:
Added a commit with some README changes. After trying to lock down Steppers for the first release, I came to the conclusion that we shouldn't do it. We may still have to do some changes for better specialization and widening conversions but if we make Steppers private, people won't be able to implement them for their own collections.

szeiger:
Merging this so we can get a release out.

PR #67, opened 2016-04-06
Add support for Scala 2.12.0-M4

szeiger:
- M4 has native support for the `JFunction` types that used to be in
  this project but they are in `scala.runtime.java8`, not
  `scala.compat.java8`, so we move them to the same location. This
  breaks compatibility with the previous release of scala-java8-compat
  but now allows cross-building between Scala 2.11 and 2.12.0-M4.
  Tests and test support code is also moved to the new package. When
  building on 2.11 we still generate our own JFunctions but on 2.12
  the native ones are used.

- JFunctions in 2.12.0-M4 are not `Serializable` (see
  https://github.com/scala/scala-dev/issues/129) so the test for that
  is disabled for the time being.

- There is currently no genjavadoc-plugin for M4, so we only build the
  docs on 2.11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ env:
 script:
   - admin/build.sh
 scala:
-  - 2.11.7
-  - 2.12.0-M3
+  - 2.11.8
+  - 2.12.0-M4
 jdk:
   - oraclejdk8
 notifications:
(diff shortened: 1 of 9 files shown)


adriaanm:
You can't publish without docs (sonatype won't let you close the repo), so we'll have to wait until we can

PR #70, opened 2016-04-18
A new approach to Stepper and Stream materialization

szeiger:
Getting the most efficient Stepper implementation requires knowledge
of the static type of a collection. This problem could be solved in the
most elegant way by integrating Steppers into the collections framework
as a replacement for Iterators (i.e. every collection gets a `stepper`
method and `iterator` delegates to `stepper` by default).

But this is at odds with the handling of specialized primitive Steppers
in the current implementation. If we want `stepper` to be an instance
methods instead of an extension method, there needs to be a single such
method for all specialized Steppers. The fundamental change in this
new implementation is to encode the translation from element type to
Stepper type (including widening conversions) as a functional dependency
via the new `StepperShape` trait.

This greatly reduces the number of implicit methods and classes and
keeps all specialized versions of `MakesStepper` together. The default
base classes support all unboxing and widening conversions so that a
simple `MakesStepper` for a collection of boxed elements only needs to
handle the `AnyStepper` case.

`keyStepper` and `valueStepper` are handled in the same way.

`StreamConverters` use a separate `StreamShape` for the translation from
element type to `BaseStream` subtype which is compatible with
`StepConverters` and supports the same primitive types and widening
conversions.

### DIFF
--- a/src/main/scala/scala/compat/java8/StreamConverters.scala
+++ b/src/main/scala/scala/compat/java8/StreamConverters.scala
@@ -292,6 +292,24 @@ with converterImpl.Priority1AccumulatorConverters
     def parStream: LongStream = seqStream.parallel
   }
 
+  implicit final class EnrichDoubleWrappedArrayWithStream(private val a: collection.mutable.WrappedArray[Double])
+    extends AnyVal with MakesSequentialStream[java.lang.Double, DoubleStream] with MakesParallelStream[java.lang.Double, DoubleStream] {
+    def seqStream: DoubleStream = java.util.Arrays.stream(a.array)
+    def parStream: DoubleStream = seqStream.parallel
+  }
+
+  implicit final class EnrichIntWrappedArrayWithStream(private val a: collection.mutable.WrappedArray[Int])
+    extends AnyVal with MakesSequentialStream[java.lang.Integer, IntStream] with MakesParallelStream[java.lang.Integer, IntStream] {
+    def seqStream: IntStream = java.util.Arrays.stream(a.array)
+    def parStream: IntStream = seqStream.parallel
+  }
+
+  implicit final class EnrichLongWrappedArrayWithStream(private val a: collection.mutable.WrappedArray[Long])
+    extends AnyVal with MakesSequentialStream[java.lang.Long, LongStream] with MakesParallelStream[java.lang.Long, LongStream] {
+    def seqStream: LongStream = java.util.Arrays.stream(a.array)
+    def parStream: LongStream = seqStream.parallel
+  }
+
   implicit val primitiveAccumulateDoubleStream = new PrimitiveStreamAccumulator[Stream[Double], DoubleAccumulator] {
     def streamAccumulate(stream: Stream[Double]): DoubleAccumulator = 
       stream.collect(DoubleAccumulator.supplier, DoubleAccumulator.boxedAdder, DoubleAccumulator.merger)
--- a/src/main/scala/scala/compat/java8/StreamConverters.scala
+++ b/src/main/scala/scala/compat/java8/StreamConverters.scala
@@ -21,7 +21,7 @@ sealed trait StreamShape[T, S <: BaseStream[_, S]] {
   def fromKeyStepper  (mk: MakesKeyValueStepper[T, _, _], par: Boolean): S
   def fromValueStepper(mk: MakesKeyValueStepper[_, T, _], par: Boolean): S
 }
-object StreamShape extends StreamShapeLowPrio {
+object StreamShape extends StreamShapeLowPriority {
   // primitive
   implicit val IntValue =    intStreamShape[Int]
   implicit val LongValue =   longStreamShape[Long]
@@ -33,7 +33,7 @@ object StreamShape extends StreamShapeLowPrio {
   implicit val CharValue    = intStreamShape[Char]
   implicit val FloatValue   = doubleStreamShape[Float]
 }
-trait StreamShapeLowPrio {
+trait StreamShapeLowPriority {
   protected[this] abstract class BaseStreamShape[T, S <: BaseStream[_, S], St <: Stepper[_]](implicit ss: StepperShape[T, St]) extends StreamShape[T, S] {
     final def fromStepper     (mk: MakesStepper[T, _],            par: Boolean): S = stream(mk.stepper,      par)
     final def fromKeyStepper  (mk: MakesKeyValueStepper[T, _, _], par: Boolean): S = stream(mk.keyStepper,   par)
--- a/src/main/scala/scala/compat/java8/StreamConverters.scala
+++ b/src/main/scala/scala/compat/java8/StreamConverters.scala
@@ -6,6 +6,57 @@ import java.util.stream._
 import scala.compat.java8.collectionImpl._
 import scala.compat.java8.converterImpl._
 
+/** Classes or objects implementing this trait create streams suitable for sequential use */
+trait MakesSequentialStream[T, SS <: java.util.stream.BaseStream[_, SS]] extends Any {
+  def seqStream: SS
+}
+
+/** Classes or objects implementing this trait create streams suitable for parallel use */
+trait MakesParallelStream[T, SS <: java.util.stream.BaseStream[_, SS]] extends Any {
+  def parStream: SS
+}
+
+sealed trait StreamShape[T, S <: BaseStream[_, S]] {
+  def fromStepper     (mk: MakesStepper[T, _],            par: Boolean): S
+  def fromKeyStepper  (mk: MakesKeyValueStepper[T, _, _], par: Boolean): S
+  def fromValueStepper(mk: MakesKeyValueStepper[_, T, _], par: Boolean): S
+}
+object StreamShape extends StreamShapeLowPrio {
+  // primitive
+  implicit val IntValue =    intStreamShape[Int]
+  implicit val LongValue =   longStreamShape[Long]
+  implicit val DoubleValue = doubleStreamShape[Double]
+
+  // widening
+  implicit val ByteValue    = intStreamShape[Byte]
+  implicit val ShortValue   = intStreamShape[Short]
+  implicit val CharValue    = intStreamShape[Char]
+  implicit val FloatValue   = doubleStreamShape[Float]
+}
+trait StreamShapeLowPrio {
+  protected[this] abstract class BaseStreamShape[T, S <: BaseStream[_, S], St <: Stepper[_]](implicit ss: StepperShape[T, St]) extends StreamShape[T, S] {
+    final def fromStepper     (mk: MakesStepper[T, _],            par: Boolean): S = stream(mk.stepper,      par)
+    final def fromKeyStepper  (mk: MakesKeyValueStepper[T, _, _], par: Boolean): S = stream(mk.keyStepper,   par)
+    final def fromValueStepper(mk: MakesKeyValueStepper[_, T, _], par: Boolean): S = stream(mk.valueStepper, par)
+    @inline private[this] def stream(st: St, par: Boolean): S = mkStream(if(par) st.anticipateParallelism else st, par)
+    protected[this] def mkStream(st: St, par: Boolean): S
+  }
+  protected[this] def intStreamShape[T](implicit ss: StepperShape[T, IntStepper]): StreamShape[T, IntStream] = new BaseStreamShape[T, IntStream, IntStepper] {
+    protected[this] def mkStream(st: IntStepper, par: Boolean): IntStream = StreamSupport.intStream(st, par)
+  }
+  protected[this] def longStreamShape[T](implicit ss: StepperShape[T, LongStepper]): StreamShape[T, LongStream] = new BaseStreamShape[T, LongStream, LongStepper] {
+    protected[this] def mkStream(st: LongStepper, par: Boolean): LongStream = StreamSupport.longStream(st, par)
+  }
+  protected[this] def doubleStreamShape[T](implicit ss: StepperShape[T, DoubleStepper]): StreamShape[T, DoubleStream] = new BaseStreamShape[T, DoubleStream, DoubleStepper] {
+    protected[this] def mkStream(st: DoubleStepper, par: Boolean): DoubleStream = StreamSupport.doubleStream(st, par)
+  }
+
+  // reference
+  implicit def anyStreamShape[T]: StreamShape[T, Stream[T]] = new BaseStreamShape[T, Stream[T], AnyStepper[T]] {
+    protected[this] def mkStream(st: AnyStepper[T], par: Boolean): Stream[T] = StreamSupport.stream(st, par)
+  }
+}
+
 trait PrimitiveStreamAccumulator[S, AA] {
   def streamAccumulate(stream: S): AA
 }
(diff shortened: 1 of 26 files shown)


szeiger:
@Ichoran, do you have time to review this?

Ichoran:
I can on Wednesday morning.

src/main/scala/scala/compat/java8/StreamConverters.scala
```diff
@@ -6,182 +6,85 @@ import java.util.stream._
...
-  }
-  implicit class EnrichBoxedShortKeySteppableWithSeqKeyStream[CC](cc: CC)(implicit steppize: CC => MakesKeyStepper[AnyStepper[Short]]) {
-    def seqKeyStream: IntStream = StreamSupport.intStream(new Stepper.WideningShortStepper(steppize(cc).keyStepper), false)
+  // widening
+  implicit val ByteValue    = intStreamShape[Byte]
+  implicit val ShortValue   = intStreamShape[Short]
+  implicit val CharValue    = intStreamShape[Char]
+  implicit val FloatValue   = doubleStreamShape[Float]
+}
+trait StreamShapeLowPrio {
```
Ichoran: Can we spell out "Priority" here?

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
```diff
@@ -5,27 +5,117 @@ import language.implicitConversions
...
+  implicit val LongValue    = valueShape[Long, LongStepper]
+  implicit val DoubleValue  = valueShape[Double, DoubleStepper]
+
+  // widening
+  implicit val ByteValue    = valueShape[Byte, IntStepper]
+  implicit val ShortValue   = valueShape[Short, IntStepper]
+  implicit val CharValue    = valueShape[Char, IntStepper]
+  implicit val FloatValue   = valueShape[Float, DoubleStepper]
+}
+trait StepperShapeLowPrio {
```
Ichoran: Again, don't think it's worth the potential loss of clarity to save the characters `rity`.

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
```diff
@@ -5,27 +5,117 @@ import language.implicitConversions
...
-  def parStream: SS
+trait MakesKeyValueStepper[K, V, +Extra] extends Any {
+  /** Generates a fresh stepper of type `S` over map keys of type `K` */
+  def keyStepper[S <: Stepper[_]](implicit ss: StepperShape[K, S]): S with Extra
+
+  /** Generates a fresh stepper of type `S` over map values of type `V` */
+  def valueStepper[S <: Stepper[_]](implicit ss: StepperShape[V, S]): S with Extra
+}
+
+sealed trait StepperShape[T, S <: Stepper[_]] { def ref: Boolean }
```
Ichoran: Can you document what `ref` does, since it's public?

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
```diff
@@ -5,27 +5,117 @@ import language.implicitConversions
...
 }
 
-trait MakesStepper[+T <: Stepper[_]] extends Any {
-  /** Generates a fresh stepper of type `T` */
-  def stepper: T
+/** Superclass for `MakesStepper` implementations which support parallelization. At least the `AnyStepper` case must be
+  * implemented, all others default to building an `AnyStepper` and putting an unboxing conversion on top. */
+trait MakesParStepper[T] extends Any with MakesStepper[T, EfficientSubstep] {
+  def stepper[S <: Stepper[_]](implicit ss: StepperShape[T, S]) = (ss match {
+    case StepperShape.IntValue    => new Stepper.UnboxingIntStepper   (stepper(StepperShape.anyStepperShape[T]).asInstanceOf[AnyStepper[Int   ]]) with EfficientSubstep
```
Ichoran: This is a pretty big case statement.  Can you add a method to StepperShape that evokes the right boxing?  I'm pretty sure that multiple dispatch is going to be faster in this case, and it'd certainly help with DRY.

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
```diff
@@ -5,27 +5,117 @@ import language.implicitConversions
...
+    case StepperShape.LongValue   => new Stepper.UnboxingLongStepper  (stepper(StepperShape.anyStepperShape[T]).asInstanceOf[AnyStepper[Long  ]]) with EfficientSubstep
+    case StepperShape.DoubleValue => new Stepper.UnboxingDoubleStepper(stepper(StepperShape.anyStepperShape[T]).asInstanceOf[AnyStepper[Double]]) with EfficientSubstep
+    case StepperShape.ByteValue   => new Stepper.UnboxingByteStepper  (stepper(StepperShape.anyStepperShape[T]).asInstanceOf[AnyStepper[Byte  ]]) with EfficientSubstep
+    case StepperShape.ShortValue  => new Stepper.UnboxingShortStepper (stepper(StepperShape.anyStepperShape[T]).asInstanceOf[AnyStepper[Short ]]) with EfficientSubstep
+    case StepperShape.CharValue   => new Stepper.UnboxingCharStepper  (stepper(StepperShape.anyStepperShape[T]).asInstanceOf[AnyStepper[Char  ]]) with EfficientSubstep
+    case StepperShape.FloatValue  => new Stepper.UnboxingFloatStepper (stepper(StepperShape.anyStepperShape[T]).asInstanceOf[AnyStepper[Float ]]) with EfficientSubstep
+    case _                        => throw new NotImplementedError("AnyStepper must be handled in `stepper` implementations")
```
Ichoran: This seems too fragile for something that users can extend.  Can you maybe have an extra abstract `anyStepper` method that this defers to (so the user must fill it in), and then at least users won't accidentally fail to realize that they need to do anything (or some particular thing) and then have it blow up at runtime?

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
```diff
@@ -5,27 +5,117 @@ import language.implicitConversions
...
+  }).asInstanceOf[S with EfficientSubstep]
 }
 
-trait MakesKeyStepper[+T <: Stepper[_]] extends Any {
-  /** Generates a fresh stepper of type `T` over map keys */
-  def keyStepper: T
+/** Superclass for `MakesStepper` implementations which do not support parallelization. At least the `AnyStepper` case must be
+  * implemented, all others default to building an `AnyStepper` and putting an unboxing conversion on top. */
+trait MakesSeqStepper[T] extends Any with MakesStepper[T, Any] {
```
Ichoran: Can we DRY this more by adding another trait that has the implementation parameterized on the return type?  AFAICT it is exactly the same code.  (Then again, if `StepperShape` handles it, maybe there won't be anything else to do.)

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
```diff
@@ -5,27 +5,117 @@ import language.implicitConversions
...
+    case StepperShape.DoubleValue => new Stepper.UnboxingDoubleStepper(keyStepper(StepperShape.anyStepperShape[K]).asInstanceOf[AnyStepper[Double]]) with EfficientSubstep
+    case StepperShape.ByteValue   => new Stepper.UnboxingByteStepper  (keyStepper(StepperShape.anyStepperShape[K]).asInstanceOf[AnyStepper[Byte  ]]) with EfficientSubstep
+    case StepperShape.ShortValue  => new Stepper.UnboxingShortStepper (keyStepper(StepperShape.anyStepperShape[K]).asInstanceOf[AnyStepper[Short ]]) with EfficientSubstep
+    case StepperShape.CharValue   => new Stepper.UnboxingCharStepper  (keyStepper(StepperShape.anyStepperShape[K]).asInstanceOf[AnyStepper[Char  ]]) with EfficientSubstep
+    case StepperShape.FloatValue  => new Stepper.UnboxingFloatStepper (keyStepper(StepperShape.anyStepperShape[K]).asInstanceOf[AnyStepper[Float ]]) with EfficientSubstep
+    case _                        => throw new NotImplementedError("AnyStepper case must be handled in `keyStepper` implementations")
+  }).asInstanceOf[S with EfficientSubstep]
+
+  def valueStepper[S <: Stepper[_]](implicit ss: StepperShape[V, S]) = (ss match {
```
Ichoran: I think this whole thing could be just `ss.unboxing(valueStepper(StepperShape.anyStepperShape[V]))`

Ichoran:
Partway through the review, but I'm out of time for right now.  I'll try to finish later tonight or, failing that, (my) Wednesday morning.

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
szeiger: This can't be in StepperShape because all implementations are slightly different (calling `stepper`, `keyStepper` or `valueStepper`, creating a wrapper with out without `EfficientSubstep`). We could potentially improve performance by dispatching on an Int value representing the StepperShape instead of using the object identities (like `StepperShape.IntValue`) so that a TABLESWITCH can be used.

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
szeiger: I tried this at first but abandoned it for some reason. I can try again.

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
szeiger: We need to create instances of different Steppers either `with EfficientSubstep` or without. There's no way to abstract over that.

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
szeiger: Hm, that could actually work (with two different methods for seq/par)

src/main/scala/scala/compat/java8/converterImpl/MakesSteppers.scala
szeiger: I tried again and now I remember why I didn't do it this way. There are types like `BitSet` for which an `AnyStepper` doesn't make sense in the current implementation because the element type is statically known to be `Int`. Consequently there is no `AnyStepper` implementation for `BitSet`. The situation would be different if we put the `stepper` methods directly into collection classes. `BitSet`'s `stepper` methods would have to take the situation into account that a `BitSet` is for example seen as a `Set[_]`, so it would have to return an `AnyStepper[_]`, but that just means that the `stepper` implementation needs to check for the type and call `super.stepper` if it's not recognized. `BitSet` would then inherit the `AnyStepper` implementation from its parent class.

szeiger: On second thought, there's no reason why `RichBitSetCanStep` would want to extend `MakesParStepper[Int]` instead of `MakesStepper[Int, EfficientSubstep]`. Either you still call `super.stepper`, then you need the `AnyStepper` version, or you don't, then you don't have to extend the abstract implementations.

szeiger:
Updated based on review comments.

src/main/scala/scala/compat/java8/StreamConverters.scala
```diff
@@ -6,182 +6,85 @@ import java.util.stream._
...
-    def seqStream: LongStream = StreamSupport.longStream(steppize(cc).stepper, false)
-  }
-  implicit class EnrichScalaCollectionWithSeqDoubleKeyStream[CC](cc: CC)(implicit steppize: CC => MakesKeyStepper[DoubleStepper]) {
-    def seqKeyStream: DoubleStream = StreamSupport.doubleStream(steppize(cc).keyStepper, false)
+trait PrimitiveStreamAccumulator[S, AA] {
+  def streamAccumulate(stream: S): AA
+}
+
+trait PrimitiveStreamUnboxer[A, S] {
+  def apply(boxed: Stream[A]): S
```
Ichoran: I'm not sure `apply` is the best choice for a method name, since it makes it impossible (with that signature) to have something that is both a function and a `PrimitiveStreamUnboxer`.  If it's for internal use only, just seal it and it's fine.  Otherwise I would pick something like `streamUnbox` to mirror `streamAccumulate` above.

src/main/scala/scala/compat/java8/converterImpl/StepsArray.scala
```diff
@@ -81,46 +70,18 @@ extends StepsLongLikeIndexed[StepsLongArray](_i0, _iN) {
...
-}
-
-final class RichArrayLongCanStep(private val underlying: Array[Long]) extends AnyVal with MakesStepper[LongStepper with EfficientSubstep] {
-  @inline def stepper: LongStepper with EfficientSubstep = new StepsLongArray(underlying, 0, underlying.length)
+final class RichArrayCanStep[T](private val underlying: Array[T]) extends AnyVal with MakesStepper[T, EfficientSubstep] {
+  override def stepper[S <: Stepper[_]](implicit ss: StepperShape[T, S]) = ((ss.shape: @switch) match {
+    case StepperShape.Reference =>
+      if(underlying.isInstanceOf[Array[Boolean]])
+                                     new StepsBoxedBooleanArray  (underlying.asInstanceOf[Array[Boolean]], 0, underlying.length)
+      else                           new StepsObjectArray[AnyRef](underlying.asInstanceOf[Array[AnyRef ]], 0, underlying.length)
```
Ichoran: I know it's not very useful to stream Unit, but will the `AnyRef` case catch it fine?  It's certainly a problem if you type something as `AnyRef` when it's actually `[Lscala.runtime.BoxedUnit`, but I think we will be read-only here so it's okay?  You at least should leave a comment explaining why you think it's okay.

Ichoran:
Aside for the few extra comments I left, this looks great!  Definitely a more compact design.

Have you run the benchmarks again to see if performance stays comparable (especially for small collections)?  There's always a concern that even though the extra abstraction is logically entirely removable that in practice it will cause a performance hit.

src/main/scala/scala/compat/java8/StreamConverters.scala
szeiger: This is from your original PR, it just got moved around. I haven't really looked into this area of the codebase yet after the initial review.

src/main/scala/scala/compat/java8/converterImpl/StepsArray.scala
szeiger: JVM arrays are covariant, so we can treat every object array as an `Array[AnyRef]` for the purpose of reading from it. Any "magical" Scala type (like `Unit` or `Null`) falls into that category, not matter if Scala treats it as a primitive or reference type. `Boolean` is the only primitive *JVM* type that gets mapped to `AnyStepper`, therefore it gets special treatment here. Since Scala, unlike the JVM, treats arrays as invariant, we do not have to cover the case where a primitive array could be seen as an object array.

szeiger:
I ran a subset of the benchmark suite (fast summation on ArrayBuffer, Array and Vector). My original PR lost a few percent of performance in some cases but I couldn't pinpoint the reason. I saw bigger losses for the large test cases where the time spent constructing the Stepper should have less of an impact.

Measurements on the updated PR aren't fully conclusive, either. The Vector test with 10 elements is still a bit slower, but the one with ArrayBuffer is faster. I'd have to run more comprehensive tests on an idle machine but so far it looks like this doesn't have a negative impact on performance compared the old version (before this PR).

As the next step I want to try removing the specialized Steppers for boxed collections. I saw a few percent of performance degradation when I tried this with the old version. We could remove quite a bit of code by always going through AnyStepper in these cases but if the manually specialized versions turn out to still be measurably faster, they may be worth keeping.

PR #74, opened 2016-08-13
adds the scala.runtime.java8 package to the published jar

### LINKED ISSUE #73: JFunction not added to the JAR
The actualy JAR won't contain the JFunction class.

At the moment this class is gone missing. It's neither on RC3 nor will it be included in the publishLocal jar.
But it will be generated. it just looks like the publish task won't add src_managed classes?

Edit:
```
OsgiKeys.exportPackage := Seq(s"scala.compat.java8.*;version=${version.value}", s"scala.runtime.java8.*;version=${version.value}"),
```
needs to be set, however it should probably include a switch for < 2.12

schmitch:
fixes #73

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,6 +8,13 @@ def jwrite(dir: java.io.File)(name: String, content: String) = {
   f
 }
 
+def osgiExport(scalaVersion: String, version: String) = {
+  (CrossVersion.partialVersion(scalaVersion) match {
+    case Some((2, 11)) => Seq(s"scala.runtime.java8.*;version=${version}")
+    case _ => Nil
+  }) ++ Seq(s"scala.compat.java8.*;version=${version}")
+}
+
 lazy val commonSettings = Seq(
   scalaVersion := "2.11.8",
   crossScalaVersions := List("2.11.8", "2.12.0-M5"),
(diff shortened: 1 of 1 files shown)


szeiger:
I think we still need to merge this for the `PackageDummy` class in 2.11.

build.sbt
```diff
@@ -8,6 +8,13 @@ def jwrite(dir: java.io.File)(name: String, content: String) = {
   f
 }
 
+def osgiExport(scalaVersion: String, version: String) = {
+  (CrossVersion.partialVersion(scalaVersion) match {
+    case Some((2, major)) if major == 11 => Seq(s"scala.runtime.java8.*;version=${version}")
```
szeiger: Why not `case Some((2, 11)) =>`?

build.sbt
```diff
@@ -62,7 +69,7 @@ lazy val root = (project in file(".")).
     }.taskValue,
 
     sourceGenerators in Compile <+= (sourceManaged in Compile, scalaVersion) map { (dir, v) =>
-      if(v.startsWith("2.11.")) {
+      if(v.startsWith("2.11")) {
```
szeiger: Are there any valid 2.11 version numbers that do not start with `2.11.`?

schmitch: Guess not

build.sbt
schmitch: I just copied that.

schmitch:
resolved the issues

szeiger:
LGTM

PR #76, opened 2016-08-29
Provide compatibility with current Scala 2.12.x (close to RC1)

szeiger:
- Add the required `override` modifiers to trait methods that now
  extend default methods of Java interfaces.

- Add casts in some of these methods because the Scala primitive types
  are not compatible at the type level with the (now inherited) Java
  boxed types.

- Use the boxed JFunction types in BoxingTest because the generic ones
  do not exist anymore.

- Tests in LambdaTest need to be commented out because usage from Java
  changed significantly after https://github.com/scala/scala/pull/5298
  but we can still cross-build the actual library against 2.11 and
  2.12.0-M5. At some point we should consider branching.


szeiger:
Used in https://github.com/scala/community-builds/pull/262. Review by @SethTisue

SethTisue:
code LGTM

"usage from Java changed significantly" — shouldn't that mean updates to README.md?

szeiger:
I'm working on restoring some of the old functionality. Hopefully the end result will only require minor changes compared to M5.

Issue #77, opened 2016-08-30
release 1.0.0 on 2.12.0


adriaanm:
- [X] fix 2.12 compat: https://github.com/scala/scala-java8-compat/pull/76
- [ ] bring back convenience that was lost with https://github.com/scala/scala/pull/5298 (we'll keep the classes in scala/scala as-is)

adriaanm:
@szeiger, could you take the lead on getting scala-java8-compat 1.0.0 out along with 2.12?

@retronym, please add your suggestions on how to proceed to my todo list above (my meeting notes were a bit sketch)

szeiger:
Moving JFunction and related classes back into scala-java8-compat: https://github.com/scala/scala-java8-compat/pull/78

schmitch:
It's also important to look into: #74 after #78. would be not cool to have 1.0.0 without `JFunction` at all.

PR #78, opened 2016-08-31
Restore backwards compatibility in Scala 2.12.0-RC1

szeiger:
- The various JFunction and JProcedure sources that were removed from
  Scala in https://github.com/scala/scala/pull/5298 are now included
  here when building for Scala 2.12.x (except M5).

- Code that relies on specialized JFunction types being subtypes of the
  generic JFunction types will inevitably break. This is unavoidable.
  The specialized types are still part of Scala but they extend Scala
  function types directly.

- All JFunctionN types and JFunction.func methods are deprecated. They
  are provided for source compatibility between 2.11 and 2.12 only.

- The JProcedureN classes, JFunction.funcSpecialized methods and
  JFunction.proc methods are still useful in 2.12 for easy
  interoperability between Scala specialization and Java’s primitive
  types from Java code.

Fixes https://github.com/scala/scala-java8-compat/issues/77. Review by @retronym.


retronym:
Does this address / overlap with #74?

build.sbt
```diff
@@ -9,8 +9,9 @@ def jwrite(dir: java.io.File)(name: String, content: String) = {
 }
 
 lazy val commonSettings = Seq(
-  scalaVersion := "2.11.8",
-  crossScalaVersions := List("2.11.8", "2.12.0-M5"),
+  resolvers += "scala-pr" at "https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/",
+  crossScalaVersions := List("2.12.0-ab61fed-SNAPSHOT", "2.11.8", "2.12.0-M5"),
```
retronym: This seems okay for now so long as we remove again after 2.12.0-RC1

szeiger:
It may cause conflicts. I delayed looking into #74 to get this done first.

retronym:
This seems to put us in a hybrid state of generated and generated-then-edited sources. Should we just check in all the generated sources at this point, and leave the generator code around as a SBT command that overwrites those sources?

retronym:
The other change to consider is serializability. As a separate PR, we could to create a set interfaces like:

```java
interface SerializableFunction1 extends Function1, Serializable
```

with a corresponding set of factory methods that use these as the parameter type.

retronym:
Aside from those details, the overall direction here LGTM. Thanks for picking this up!

szeiger:
Currently all JFunctions extend Serializable, to the problem doesn't arise if you use those. If I understand the original motivation for this correctly, removing JFunctions would not be problem. People can always declare their own functional interfaces that implement Serializable, as long as the Scala compiler knows how to deal with them.

szeiger:
How was the code generated? It would be nice to run the generator for JFunction and JProcedure classes (like we do for other classes) instead of keeping the source around. I copied them from the Scala repo (before they were removed) but I don't see any generator code there.

szeiger:
Never mind, found it (after rewriting it from scratch...)

szeiger:
Updated with source generator.

retronym:
I think we should also restore all the types into `scala.compat.java8` (reverting https://github.com/scala/scala-java8-compat/commit/55df152d044be03cd528790695f98991e5e85771#diff-a23da058b34325a664d3a84acd3da78aL28).

This will avoid the "split packages", which are discourages in OSGi and will be incompatible with Java 9's module system.

retronym:
```
sbt> +compile
...
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: com.typesafe.genjavadoc#genjavadoc-plugin_2.12.0-ab61fed-SNAPSHOT;0.10: not found
[warn] 	:: org.scala-lang#scala-compiler;2.12.0-ab61fed-SNAPSHOT: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn]
[warn] 	Note: Unresolved dependencies path:
[warn] 		com.typesafe.genjavadoc:genjavadoc-plugin_2.12.0-ab61fed-SNAPSHOT:0.10 (/Users/jason/code/scala-java8-compat/build.sbt#L113)
[warn] 		  +- org.scala-lang.modules:scala-java8-compat_2.12.0-ab61fed-SNAPSHOT:0.8.0-SNAPSHOT
[warn] 		org.scala-lang:scala-compiler:2.12.0-ab61fed-SNAPSHOT
[warn] 		  +- org.scala-lang.modules:scala-java8-compat_2.12.0-ab61fed-SNAPSHOT:0.8.0-SNAPSHOT
```

retronym:
```
[warn] ==== scala-pr: tried
[warn]   https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/org/scala-lang/scala-compiler/2.12.0-ab61fed-SNAPSHOT/scala-compiler-2.12.0-ab61fed-SNAPSHOT.pom

[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: org.scala-lang#scala-compiler;2.12.0-ab61fed-SNAPSHOT: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn]
[warn] 	Note: Unresolved dependencies path:
[warn] 		org.scala-lang:scala-compiler:2.12.0-ab61fed-SNAPSHOT
[warn] 		  +- org.scala-lang.modules:scala-java8-compat_2.12.0-ab61fed-SNAPSHOT:0.8.0-SNAPSHOT
```

I don't see https://github.com/scala/scala/commit/ab61fed in the PR snapshot repo.

szeiger:
How long do we keep PR snapshot builds? This was available last week. Is a nightly build a better choice?

I assumed split packages would work in Java 9 the same way they do in OSGi. If this is not the case, we'll have to do another breaking change :-(

szeiger:
Updated with the new old packages (and a new nightly build that still exists in the repo)

SethTisue:
> How long do we keep PR snapshot builds?

a long time, maybe indefinitely, judging from a glance at https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/org/scala-lang/scala-compiler/

> This was available last week. Is a nightly build a better choice

we keep the nightlies for 6 months (as per recent emails with @fsalvi)

adriaanm:
Regarding the artifact, a search for *ab61fed* at https://scala-ci.typesafe.com/artifactory/ reveals the library and reflect jars are there, but not the compiler. Fluke during a build?

adriaanm:
The fluke was me -- I aborted (accidentally, I think) https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-validate-publish-core/3414/console

retronym:
LGTM

PR #81, opened 2016-10-18
Bump to Scala 2.12.0-RC2

retronym:


retronym:
I've worked around the lack of a genjavadoc build for RC2 by using the RC1 version instead.

PR #84, opened 2016-10-29
Bump to new snapshot version

retronym:


retronym:
@Ichoran @szeiger I think we're at a point were we should publish 1.0.0 of this module and commit to backwards binary compatibility, which would help upstream projects. WDYT? It would be advantageous to do this relatively soon as upstream libraries will be releasing for 2.12.0 soon.

Ichoran:
I recommend running something like the collections-laws tests on the library before considering it 1.0.  I wrote so much new code that wasn't very heavily tested (everything had basic tests, at least) that I am sure there are bugs in there waiting to be found.  I don't need to be the one to run these tests, but it would be good if someone did.  I am giving an important talk on Nov. 4, after which I could do it.

In particular, with iterators, it's important to test all the variant ways that an underlying collection can be sized--e.g. all the different sizes of immutable maps that actually have distinct implementations--and if there are any cases where you're modifying and wrapping (e.g. flatMap, collect, etc.), to test the functionality on both plain and modified/wrapped cases.

retronym:
Okay, I'll keep this pre-1.0 for now.

PR #88, opened 2017-04-25
Upate scala-module-plugin, updates to tag-based publishing

lrytz:
After this patch, `.travis.yml` no longer defines any Scala versions,
this is always done in sbt. This way there's a single place to update
Scala versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: scala
 sudo: false
 env:
   global:
-    - PUBLISH_JDK=oraclejdk8
     # PGP_PASSPHRASE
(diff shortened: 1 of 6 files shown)


lrytz:
For the 2.13.0-M1 cross-build, we need `genjavadoc-plugin_2.13.0-M1` first

lrytz:
This is ready now, genjavadoc was published and the build is green.

SethTisue:
released 0.8.0 for 2.13.0-M1: https://travis-ci.org/scala/scala-java8-compat/builds/231264671

PR #89, opened 2017-06-02
Optimise FutureConverters.toScala

t3hnar:
Optimise `FutureConverters.toScala` for case when `CompletionStage` backed by `CompletableFuture`

### DIFF
--- a/src/main/scala/scala/compat/java8/FutureConverters.scala
+++ b/src/main/scala/scala/compat/java8/FutureConverters.scala
@@ -4,10 +4,11 @@
 package scala.compat.java8
 
 import scala.concurrent.java8.FuturesConvertersImpl._
-import scala.concurrent.{ Future, Promise, ExecutionContext, ExecutionContextExecutorService, ExecutionContextExecutor, impl }
-import java.util.concurrent.{ CompletionStage, Executor, ExecutorService, CompletableFuture }
-import scala.util.{ Try, Success, Failure }
-import java.util.function.{ BiConsumer, Function ⇒ JF, Consumer, BiFunction }
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, ExecutionContextExecutorService, Future, Promise}
+import java.util.concurrent.{CompletableFuture, CompletionStage, ExecutionException, Executor, ExecutorService}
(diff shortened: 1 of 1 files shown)


SethTisue:
@t3hnar I'm sorry, but this seems to have fallen through the cracks. now it needs a rebase, because of merge conflicts in the `import`s

t3hnar:
@SethTisue done

SethTisue:
the compilation failures are 2.13-only and unrelated to this PR, see https://github.com/scala/scala-java8-compat/issues/97

SethTisue:
now we need a reviewer, maybe @retronym?

SethTisue:
review by @viktorklang?

src/main/scala/scala/compat/java8/FutureConverters.scala
```diff
@@ -78,6 +79,8 @@ object FutureConverters {
   def toScala[T](cs: CompletionStage[T]): Future[T] = {
     cs match {
       case cf: CF[T] => cf.wrapped
+      case cf: CompletableFuture[T] if cf.isDone =>
+        Future fromTry Try(cf.get()).recoverWith { case e: ExecutionException => Failure(e.getCause) }
```
viktorklang: Given this code it is not clear to me that this is in fact an optimization. Is there any benchmark performed that shows what improvement this change leads to?

t3hnar: @viktorklang Might be I missing something, but I believe that optimised version works faster because of removal of possible context switch

viktorklang:
*@t3hnar* which benchmark shows what improvement?
-- 
Cheers,
√

t3hnar:
@viktorklang there is no benchmark, just a blind believe, so feel free to reject this pr.

src/main/scala/scala/compat/java8/FutureConverters.scala
viktorklang: @t3hnar I suspect this to be slower: there will be one allocation for the thunk to Try.apply, then another thunk for the recoverWith function. For the other code it will be a single allocation of the P.

PR #91, opened 2017-07-24
in build, avoid method deprecated in sbt 0.13.16

SethTisue:
toError is deprecated now. the replacement code is as
recommended by the deprecation message


SethTisue:
not actually upgrading to sbt 0.13.16 yet since it's still in RCs, but since we have `-Xfatal-warnings` enabled for the build definition, this change is required for testing the community build with 0.13.16

Issue #97, opened 2018-01-24
compilation failures on tests in Scala 2.13 community build


SethTisue:
from https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-community-build/905/consoleFull : 

```
[scala-java8-compat] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.9/project-builds/scala-java8-compat-3b0fd4decb75f5e349e7c755590b4855660ac3b4/src/test/scala/scala/compat/java8/StreamConvertersTest.scala:249: type mismatch;
[scala-java8-compat] [error]  found   : java.util.stream.Stream[Short]
[scala-java8-compat] [error]  required: java.util.stream.IntStream
[scala-java8-compat] [error]     assertEquals(Vector[Short](1.toShort, 2.toShort, 3.toShort), (Array[Short](1.toShort, 2.toShort, 3.toShort).seqStream: IntStream).toScala[Vector])
[scala-java8-compat] [error]                                                                                                                 ^
[scala-java8-compat] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.9/project-builds/scala-java8-compat-3b0fd4decb75f5e349e7c755590b4855660ac3b4/src/test/scala/scala/compat/java8/StreamConvertersTest.scala:253: type mismatch;
[scala-java8-compat] [error]  found   : java.util.stream.Stream[Int]
[scala-java8-compat] [error]  required: java.util.stream.IntStream
[scala-java8-compat] [error]     assertEquals(Vector[Int](1, 2, 3), (Vector[Int](1, 2, 3).seqStream: IntStream).toScala[Vector])
[scala-java8-compat] [error]                                                              ^
[scala-java8-compat] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.9/project-builds/scala-java8-compat-3b0fd4decb75f5e349e7c755590b4855660ac3b4/src/test/scala/scala/compat/java8/StreamConvertersTest.scala:254: type mismatch;
[scala-java8-compat] [error]  found   : java.util.stream.Stream[Short]
[scala-java8-compat] [error]  required: java.util.stream.IntStream
[scala-java8-compat] [error]     assertEquals(Vector[Short](1.toShort, 2.toShort, 3.toShort), (Vector[Short](1.toShort, 2.toShort, 3.toShort).seqStream: IntStream).toScala[Vector])
[...]

PR #100, opened 2018-02-22
Remove override of `Future.onSuccess`

### LINKED ISSUE #99: Overriding `Future.onSuccess` in `FutureConvertersImpl`
`FutureConvertersImpl.P` overrides `onSuccess` [here](https://github.com/scala/scala-java8-compat/blob/35831f444ca3acb214c9a276867d3a1e06d1a940/src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala#L90) with a call to `super`. Is there an important reason for that, or can it be removed?

This comes up because `onSuccess` is deprecated, and may be removed in 2.13 (see scala/scala#6319).

NthPortal:
`Future.onSuccess` is deprecated and slated for removal. See also scala/scala#6319.

Fixes #99

### DIFF
--- a/src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
+++ b/src/main/scala/scala/concurrent/java8/FutureConvertersImpl.scala
@@ -87,8 +87,6 @@ object FuturesConvertersImpl {
   }
 
   class P[T](val wrapped: CompletionStage[T]) extends DefaultPromise[T] with BiConsumer[T, Throwable] {
-    override def onSuccess[U](pf: PartialFunction[T, U])(implicit executor: ExecutionContext): Unit = super.onSuccess(pf)
-
     override def accept(v: T, e: Throwable): Unit = {
       if (e == null) complete(Success(v))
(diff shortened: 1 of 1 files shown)


NthPortal:
The build failure seems to have nothing to do with the changes from this PR - not sure what to do about that

NthPortal:
It looks like it's failing with the same error as [the latest build on master](https://travis-ci.org/scala/scala-java8-compat/builds/342406389), so maybe it's not a problem?

SethTisue:
@retronym I'm not familiar with this code. will removing the override harm us in 2.12? is this a case where we will need to have separate `src/main/scala-2.12` and `src/main/scala-2.13` directories?

SethTisue:
> It looks like it's failing with the same error as the latest build on master, so maybe it's not a problem?

correct, it's #97

SethTisue:
thanks Nth, appreciate your diligence in following through with this in multiple repos

NthPortal:
@SethTisue No problem!

PR #101, opened 2018-03-22
support building on Java 9 and 10

SethTisue:
because we're trying to get the Scala 2.12 community build up
and running on Java 9 and beyond


SethTisue:
the Travis failure is because of #97 which we expect has been fixed for 2.13.0-M4

PR #109, opened 2018-06-13
Compatibility with 2.13 collections

lrytz:
I did not make an attempt to have a single branch that cross-builds with 2.13 and 2.12/2.11. I think the effort would be too big in this case. Instead, I think we should create a new `master-2.13` branch (or similar) and merge changes from master forward.

I'm creating a PR against master anyway, so we have a place to review the changes.

genjavadoc is currently disbabled, its 2.13.0-M4 version is waiting for scalatest (https://github.com/lightbend/genjavadoc/pull/141).

This PR doesn't test / add support for new collection types yet, in particular
  - `immutable.ArraySeq`
  - `mutable.ArrayDeque`
  - `immutable.ChampHashMap` / `immutable.ChampHashSet`
  - others?


src/main/scala/scala/compat/java8/collectionImpl/Stepper.scala
```diff
@@ -167,8 +167,8 @@ trait StepperLike[@specialized(Double, Int, Long) A, +CC] { self: CC =>
   }
 
   /** Returns a Scala collection of the type requested. */
-  def to[Coll[_]](implicit cbf: collection.generic.CanBuildFrom[Nothing, A, Coll[A]]): Coll[A] = {
-    val b = cbf()
+  def to[Coll[_]](implicit facotry: collection.Factory[A, Coll[A]]): Coll[A] = {
```
retronym: typo

retronym:
I'm not familiar with this code, @szeiger and/or @Ichoran were the authors. I don't like that we still need to break the encapsulation of the collections, that suggests the `Stepper` infrastructure should be in the standard library itself. But that's in existing problem, not one introduced by this PR.

Ichoran:
@retronym - I don't and didn't like it either, and there was a plan to expose more of the internal functionality through some mechanism, but I never got that done (and wasn't entirely sure it was a good idea either).

However, if ever we are going to make `Stepper` internal to the library, now would be the time!

lrytz:
That sounds worthwile to investigate for sure.

In the interest of unblocking the ecosystem (akka), I think it would be good not to block this PR though. Other than that, I think it's ready. IMO it would be OK to do a release for M4 without documentation (since genjavadoc is not done yet). Or I could just release genjavadoc by disabling its tests on M4 (as it is missing scalatest).

lrytz:
Pushed https://github.com/scala/scala-java8-compat/tree/2.13.x

lrytz:
Released 0.9.0 for M4 https://github.com/scala/scala-java8-compat/releases/tag/v0.9.0%232.13.0-M4

PR #112, opened 2018-06-18
fix compile errors in latest Scala 2.13

xuwei-k:
fix https://github.com/scala/scala-java8-compat/issues/110

- https://github.com/scala/scala/commit/1e21c918361b22d66575acb82b57d532823b198a#diff-c4d00ae1df2a449ae3d28a7a4469e28a
- https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/1187/consoleFull

```
[scala-java8-compat] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.12/project-builds/scala-java8-compat-e21fca9479a0769dd077af3a1d088d2f68798316/src/main/scala/scala/compat/java8/converterImpl/StepConverters.scala:19: type mismatch;
[scala-java8-compat] [error]  found   : Array[_$2] where type _$2
[scala-java8-compat] [error]  required: Array[A]
[scala-java8-compat] [error]  Note: implicit method richArraySeqCanStep is not applicable here because it comes after the application point and it lacks an explicit result type
[scala-java8-compat] [error]   implicit def richArraySeqCanStep[A](underlying: collection.mutable.ArraySeq[A]) = new RichArrayCanStep[A](underlying.array)
[scala-java8-compat] [error]                                                                                                                        ^
[scala-java8-compat] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.12/project-builds/scala-java8-compat-e21fca9479a0769dd077af3a1d088d2f68798316/src/main/scala/scala/compat/java8/StreamConverters.scala:200: type mismatch;
[scala-java8-compat] [error]  found   : Array[_$2] where type _$2
[scala-java8-compat] [error]  required: Array[Double]
[scala-java8-compat] [error]  Note: implicit value accumulateLongStepper is not applicable here because it comes after the application point and it lacks an explicit result type
[scala-java8-compat] [error]     def seqStream: DoubleStream = java.util.Arrays.stream(a.array)
[scala-java8-compat] [error]                                                             ^
[scala-java8-compat] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.12/project-builds/scala-java8-compat-e21fca9479a0769dd077af3a1d088d2f68798316/src/main/scala/scala/compat/java8/StreamConverters.scala:206: type mismatch;
[scala-java8-compat] [error]  found   : Array[_$2] where type _$2
[scala-java8-compat] [error]  required: Array[Int]
[...]

PR #115, opened 2018-07-11
adapt to post 2.13.0-M4 changes

SethTisue:
~~three sorts of changes here:~~
* adding explicit result types to implicits.  not sure what Scala
  change made these fail to compile otherwise, but they are good
  changes regardless
~~* changing `.array` to `.toArray` when we are dealing with
  `ArraySeq`, because scala/scala#6837~~
~~* adding `: ClassTag` in one place, also because scala/scala#6837.
  this change is not binary compatible and isn't fully source
  compatible either, but it's okay, this is a 2.13-only branch.~~

### DIFF
--- a/src/main/scala/scala/compat/java8/StreamConverters.scala
+++ b/src/main/scala/scala/compat/java8/StreamConverters.scala
@@ -221,52 +221,58 @@ with converterImpl.Priority1AccumulatorConverters
     def parStream: LongStream = seqStream.parallel
   }
 
-  implicit val primitiveAccumulateDoubleStream = new PrimitiveStreamAccumulator[Stream[Double], DoubleAccumulator] {
-    def streamAccumulate(stream: Stream[Double]): DoubleAccumulator = 
-      stream.collect(DoubleAccumulator.supplier, DoubleAccumulator.boxedAdder, DoubleAccumulator.merger)
-  }
-  
-  implicit val primitiveAccumulateDoubleStream2 =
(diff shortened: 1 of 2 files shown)


SethTisue:
@lrytz I'm hoping you'll actually think about this a bit... I'm just somewhat hurriedly trying to get this green in 2.13 community build green since it's blocking downstream stuff

src/main/scala/scala/compat/java8/StreamConverters.scala
```diff
@@ -197,68 +197,74 @@ with converterImpl.Priority1AccumulatorConverters
...
     extends AnyVal with MakesSequentialStream[Double, DoubleStream] with MakesParallelStream[Double, DoubleStream] {
-    def seqStream: DoubleStream = java.util.Arrays.stream(a.array)
+    def seqStream: DoubleStream = java.util.Arrays.stream(a.toArray)
     def parStream: DoubleStream = seqStream.parallel
   }
 
   implicit final class EnrichIntArraySeqWithStream(private val a: collection.mutable.ArraySeq[Int])
     extends AnyVal with MakesSequentialStream[Int, IntStream] with MakesParallelStream[Int, IntStream] {
-    def seqStream: IntStream = java.util.Arrays.stream(a.array)
+    def seqStream: IntStream = java.util.Arrays.stream(a.toArray)
```
xuwei-k: https://github.com/scala/scala-java8-compat/pull/112#discussion_r195985858

lrytz:
I merged #112, so this PR can lose some weight :)

PR #117, opened 2018-07-28
Fixed NPE in VectorStepper due to dirty display vector.

Ichoran:
(Could occur with structurally shared vectors.)

Fixed by creating an iterator, which creates a clean display.  (Would probably be more efficient for everyone if this was handled a different way, but this is the easiest to code.)

### DIFF
--- a/src/main/java/scala/compat/java8/runtime/CollectionInternals.java
+++ b/src/main/java/scala/compat/java8/runtime/CollectionInternals.java
@@ -5,12 +5,19 @@ package scala.compat.java8.runtime;
 public class CollectionInternals {
     public static <A> Object[] getTable(scala.collection.mutable.FlatHashTable<A> fht) { return fht.hashTableContents().table(); }
     public static <A, E extends scala.collection.mutable.HashEntry<A,E>> scala.collection.mutable.HashEntry<A, E>[] getTable(scala.collection.mutable.HashTable<A,E> ht) { return ht.hashTableContents().table(); }
+    public static <A> boolean getDirt(scala.collection.immutable.Vector<A> v) { return v.dirty(); }
     public static <A> Object[] getDisplay0(scala.collection.immutable.Vector<A> v) { return v.display0(); }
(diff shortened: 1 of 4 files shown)


SethTisue:
review by @joshlemer?

src/main/java/scala/compat/java8/runtime/CollectionInternals.java
```diff
@@ -5,12 +5,19 @@
...
     public static <A> Object[] getDisplay1(scala.collection.immutable.Vector<A> v) { return v.display1(); }
+    public static <A> Object[] getDisplay1(scala.collection.immutable.VectorIterator<A> p) { return p.display1(); }
     public static <A> Object[] getDisplay2(scala.collection.immutable.Vector<A> v) { return v.display2(); }
+    public static <A> Object[] getDisplay2(scala.collection.immutable.VectorIterator<A> p) { return p.display2(); }
     public static <A> Object[] getDisplay3(scala.collection.immutable.Vector<A> v) { return v.display3(); }
+    public static <A> Object[] getDisplay3(scala.collection.immutable.VectorIterator<A> p) { return p.display3(); }
     public static <A> Object[] getDisplay4(scala.collection.immutable.Vector<A> v) { return v.display4(); }
+    public static <A> Object[] getDisplay4(scala.collection.immutable.VectorIterator<A> p) { return p.display4(); }
     public static <A> Object[] getDisplay5(scala.collection.immutable.Vector<A> v) { return v.display5(); }
+    public static <A> Object[] getDisplay5(scala.collection.immutable.VectorIterator<A> p) { return p.display5(); }
```
joshlemer: Could all these duplicates just take a `scala.collection.immutable.VectorPointer<A>`?

joshlemer: oh I see 
>```// (Manually specialized this way for speed.)```

Ichoran: Yeah, it cuts off about 20% of the runtime in my ad-hoc testing.

PR #118, opened 2018-08-29
Update FutureConverters.scala

etay1283:
adding a Java Future to Scala Future conversion

### DIFF
--- a/src/main/scala/scala/compat/java8/FutureConverters.scala
+++ b/src/main/scala/scala/compat/java8/FutureConverters.scala
@@ -198,4 +198,16 @@ object FutureConverters {
      */
     def toScala: Future[T] = FutureConverters.toScala(__self)
   }
+  
+  /**
+   * Wrap a Java Future result in Scala Future using implicit ExecutionContext
+   *
+   * @return a Scala Future with the Java Future result
+   *
+   */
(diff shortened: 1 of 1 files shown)


SethTisue:
review by @viktorklang?

src/main/scala/scala/compat/java8/FutureConverters.scala
```diff
@@ -198,4 +198,16 @@ object FutureConverters {
...
+  
+  /**
+   * Wrap a Java Future result in Scala Future using implicit ExecutionContext
+   *
+   * @return a Scala Future with the Java Future result
+   *
+   */
+  def toScala[T](f: java.util.concurrent.Future[T])(implicit ec: ExecutionContext): Future[T] = {
+    Future {
+      f.get()
```
viktorklang: Apologies for sounding negative, but:

This is a very dangerous conversion—not only does it not mention in the documentation, nor the method name, that it will block threads, but it also does not use the `blocking {}`-construct to let the underlying ExecutionContext to attempt to maintain liveness. Furthermore, it does not specify any timeout for the `get`-operation which means that it could tie up a thread indefinitely.

All this together, I think that this proposal is too dangerous and looks rather "innocent" from a caller-perspective.

I hope this makes sense,

√

PR #122, opened 2018-09-20
use .linesIterator to work around scala/bug#11125

SethTisue:
this incurs a deprecation warning on 2.12.6, but we don't have
fatal warnings enabled in this repo. the deprecation warnings
will go away once we start building on 2.12.7.

references scala/bug#11125


SethTisue:
this is so we can build on JDK 11 in the Scala community build

SethTisue:
merged onto 2.13.x branch in 1bcf341144e875fadcd5f267fdebb7de341779e7

PR #136, opened 2019-03-14
Fix the DurationConverters import in README

filosganga:
Fix the Duration examples typo in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ Example of conversions from the Java type ways:
 
 ```scala
 import scala.concurrent.duration._
-import scala.compat.java8.DurationConverters
+import scala.compat.java8.DurationConverters._
 
 val javaDuration: java.time.Duration = 5.seconds.toJava
 val finiteDuration: FiniteDuration = javaDuration.toScala


README.md
```diff
@@ -268,15 +268,15 @@ Example of conversions from the Java type ways:
...
+import scala.compat.java8.DurationConverters._
 
 val javaDuration: java.time.Duration = 5.seconds.toJava
 val finiteDuration: FiniteDuration = javaDuration.toScala
 ```
 
 From Java:
 ```java
-import scala.compat.java8.DurationConverters;
+import scala.compat.java8.DurationConverters._;
```
SethTisue: isn't the Java syntax something else? like `*` instead of `_`? and should it be `import static` perhaps?

README.md
```diff
@@ -268,15 +268,15 @@ Example of conversions from the Java type ways:
 
 ```scala
 import scala.concurrent.duration._
-import scala.compat.java8.DurationConverters
+import scala.compat.java8.DurationConverters._
```
SethTisue: 👍

README.md
filosganga: Ooops yes, it should be:

```
import static scala.compat.java8.DurationConverters.*;
```

filosganga: I have fixed and rebased.

filosganga: Actually, the java import was correct as is, I have reverted back, sorry.

SethTisue:
thanks!

PR #139, opened 2019-04-11
Adapt to 2.13.0-RC1

lrytz:
Includes a merge from `master` to `2.13.x`.

The idea is to leave java8-compat for 2.13 the same as for 2.12, except for Stream converters.

So java8-compat continues to provide its own implementation of `FutureConverters`, `DurationConverters`, `OptionConverters` and `FunctionConverters`, with its own wrapper classes.

For Stream converters, the situation is slightly different because they build on Steppers, which are now built into the core of the collections: `IterableOnce` has a `stepper` method, and the implementations for specific collection types now live in the standard library. Duplicating everything in java8-compat would be a big maintenance burden.

The `scala.compat.java8.StreamConverters` object still exists, but it provides a slightly different API than in 2.12-java8-compat.  The idea is to make as much of the 2.12-java8-compat API available as practicable, but use the implemenations in the 2.13 standard library.


lrytz:
Updated the description. Review by @Ichoran, @SethTisue

SethTisue:
let's see if we get any useful feedback from the community build: https://github.com/scala/community-builds/commit/3c3c57ed636bc00062fdf0bbd4cc8289b1d70ae2

johanandren:
Ran through the Akka test suite with a snapshot of this and it looks good!

lrytz:
Cool :-) did you have to do any changes, or is the API source-compatible enought for how it's used in Akka?

johanandren:
Seems source compatible! :+1:

Ichoran:
I don't think I'm going to have time to review this in worthwhile detail.

lrytz:
@Ichoran no problem. Any thoughts on the general strategy outlined in the PR description?

The changes are not actually that huge, let me summarize
  - Removed all Stepper and Accumulator implementations in this repo. Also removed helpers like `StreamShape`, `MakesSequentialStream`
  - Created type aliases for Accumulator and Stepper types in the package where they used to live
  - Remoed the `stepper` extension method
  - Created extension methods in `StepperExtensions` to bring most of the `Stepper` API back (`.accumulate`, operations like `.find`)
  - Kept the `seqStream`, `parStream`, `seqKeyStream`, `seqValueStream` extension methods we use different names in RC1 (`asJavaSeqStream` etc)
    - The implementations are the same as in RC1, so they use `StreamShape` and `StepperShape` from the standard library
  - Kept `.accumulate`, `.accumulatePrimitive`, `.toScala` extension methods for streams

Ichoran:
As far as I have time to understand it, which is probably not far enough to be of much use, the strategy sounds reasonable.  (I still am not entirely convinced that `.asJavaSeqStream` is a good name, but I guess that ship has sailed.)  Providing extension methods to allow `Stepper` to do something is good; not having those vs. having those could (if anyone used them) have been one of the biggest challenges to upgrading and/or dual-compiling.

johanandren:
FYI we only use `FunctionConverters`, `OptionConverters` and `FutureConverters`, so for enabling Akka to move forward with 2.13.0-RC1 the rest doesn't matter.

lrytz:
I think we can go ahead, maybe review by @SethTisue?

lrytz:
Release is on maven central http://repo1.maven.org/maven2/org/scala-lang/modules/scala-java8-compat_2.13.0-RC1/0.9.0

Issue #145, opened 2019-05-20
Publish for 2.13.0-RC2


raboof:
See also #139

lrytz:
I'll take a look today. I'd like to fix https://github.com/scala/scala-java8-compat/issues/144 for the RC2 release though. Not sure if I need to change the version number, or using 0.9.0 again would be ok :)

raboof:
:+1: - pushed my wip as #146 - haven't checked it really makes sense though, I just "followed the types"

lrytz:
We'll need genjavadoc as well

raboof:
https://github.com/lightbend/genjavadoc/pull/173

lrytz:
The release is on staging https://oss.sonatype.org/content/repositories/staging/org/scala-lang/modules/scala-java8-compat_2.13.0-RC2, I released the staging repo, and it's no longer in the list on nexus. but didn't show up on central yet (http://repo1.maven.org/maven2/org/scala-lang/modules/scala-java8-compat_2.13.0-RC2).

lrytz:
ah, the missing trailing slash was the problem.. http://repo1.maven.org/maven2/org/scala-lang/modules/scala-java8-compat_2.13.0-RC2/

PR #146, opened 2019-05-20
Changes for 2.13.0-RC2

raboof:
Refs #145

### DIFF
--- a/src/main/java/scala/compat/java8/ScalaStreamSupport.java
+++ b/src/main/java/scala/compat/java8/ScalaStreamSupport.java
@@ -15,7 +15,7 @@ package scala.compat.java8;
 import java.util.stream.*;
 
 import scala.collection.*;
-import scala.jdk.StreamConverters;
+import scala.jdk.javaapi.StreamConverters;
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/scala/compat/java8/ScalaStreamSupport.java
```diff
@@ -183,7 +183,7 @@ public static DoubleStream doubleStreamValues(Map<?, Double> coll) {
      * @return     A DoubleStream view of the collection which, by default, executes sequentially.
      */
     public static DoubleStream doubleStreamAccumulated(IterableOnce<Double> coll) {
-        return StreamConverters.asJavaSeqDoubleStream(scala.jdk.DoubleAccumulator$.MODULE$.fromSpecific((IterableOnce<Object>)(Object)coll));
+        return StreamConverters.asJavaSeqDoubleStream(coll);
```
lrytz: we should keep the double accumulator here. but in any case, i already had a draft pr with these changes: https://github.com/scala/scala-java8-compat/pull/143. so closing this one. sorry..!

Issue #144, opened 2019-05-14
StreamConverters.toScala on 2.13 uses value param instead of type param


lrytz:
This prevents cross-building projects using java8-compat on 2.12 and 2.13, which is the only reason for having a 2.13 release. The `toScala` method should have the same API.

PR #147, opened 2019-05-20
StreamConverters.toScala in 2.12-style

lrytz:
Fixes https://github.com/scala/scala-java8-compat/issues/144


lrytz:
Review by @dwijnand - you had a case that was affected, no?

dwijnand:
In comparing the test case in 2.13.x with the one in master there are a few changes, particularly these kinds of ones:

```scala
-      assertEq(vecO, newStream(n).accumulate.to[Vector], s"stream $n to vector")
-      assertEq(vecO, accO.to[Vector], s"stream $n to vector in parallel")
+      assertEq(vecO, newStream(n).accumulate.to(Vector), s"stream $n to vector")
+      assertEq(vecO, accO.to(Vector), s"stream $n to vector in parallel")
```

See https://github.com/scala/scala-java8-compat/compare/master...lrytz:toScalaTypeParam#diff-fb9126fd878d17950f0dbccaed6d630c.

lrytz:
Thanks for checking. The problem here is that the `Accumulator` type, when using scala-java8-compat on 2.13, is the one from the 2.13 standard library. So its `.to` method has the new 2.13 shape, it takes a (non-implicit) factory value parameter.

I'm not sure how we could bridge that gap here, i.e., I don't see a way to make the `accumulator.to[Vector]` syntax work when using scala-java8-compat on 2.13. However, one could use scala-collection-compat to make the `accumulator.to(Vector)` syntax work when using scala-java8-compat on 2.12. (scala-collection-compat doesn't add a new `to` extension method for that, but instead provides implicit conversions from collection companions to `CanBuildFrom` instances, so that the 2.12 `to` method can be used in the 2.13 style).

Issue #151, opened 2019-06-10
Automatic upcast in FutureConverters


TisonKun:
https://github.com/scala/scala-java8-compat/blob/62d4f80cc18d47aa32a08c8711d67565cf00d743/src/main/scala/scala/compat/java8/FutureConverters.scala#L66

```scala
def toJava[T, U <: T ](f: Future[U]): CompletionStage[T] = { 
```

PR #152, opened 2019-06-10
Automatic upcast in FutureConverters

TisonKun:
ref #151 

straight forward


lrytz:
Thanks for the PR, @TisonKun. I'm not sure about doing this only here, we'd also have to look at OptionConverters, StreamConverters. Note that Scala 2.13 ships with these converters in the standard library (in `scala.jdk`), so we'd have to do it there as well, also for CollectionConverters. This looks like quite a substantial change, we'd have to test it thoroughly to make sure inference (also javac's) doesn't trip up..

So I'd say that the change in this PR is too isolated to be merged. Leaving it open for now for further discussion.

TisonKun:
@lrytz Thanks for your information. Agree with leaving for further discussion. I'd learn these converters  things and find a way to cover all topics and tests if possible.

PR #211, opened 2021-03-05
declare versionScheme and enable MiMa

### LINKED ISSUE #210: Build should declare a `versionScheme`
as per https://scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html and discussion at scala/sbt-scala-module#111

SethTisue:
fixes #210

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ language: scala
 
 scala:
   - 2.11.12
-  - 2.12.12
-  - 2.13.4
+  - 2.12.13
+  - 2.13.5
 
 env:
   - ADOPTOPENJDK=8
--- a/build.sbt
+++ b/build.sbt
@@ -78,6 +78,19 @@ lazy val scalaJava8Compat = (project in file("."))
     // think about going 1.0, it's been a while
     scalaModuleMimaPreviousVersion := Some("0.9.1"),
 
+    mimaBinaryIssueFilters ++= {
+      import com.typesafe.tools.mima.core._, ProblemFilters._
+      Seq(
+        // bah
+        exclude[IncompatibleSignatureProblem]("*"),
+        // mysterious -- see scala/scala-java8-compat#211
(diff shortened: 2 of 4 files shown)


SethTisue:
MiMa failures I'll need to dig into

SethTisue:
okay so the MiMa failures are only on 2.11 and 2.12, not on 2.13.

```
[error]  * method enrichAsJavaIntFunction(scala.Function1)scala.Function1 in trait scala.compat.java8.Priority1FunctionConverters does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.compat.java8.Priority1FunctionConverters.enrichAsJavaIntFunction")
[error]  * method enrichAsJavaIntFunction(scala.Function1,scala.Predef#=:=)scala.Function1 in trait scala.compat.java8.Priority1FunctionConverters is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.compat.java8.Priority1FunctionConverters.enrichAsJavaIntFunction")
[error]  * method enrichAsJavaIntFunction(scala.Function1)scala.Function1 in object scala.compat.java8.FunctionConverters.package does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.compat.java8.FunctionConverters.package.enrichAsJavaIntFunction")
[error]  * static method enrichAsJavaIntFunction(scala.Function1)scala.Function1 in class scala.compat.java8.FunctionConverters.package does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.compat.java8.FunctionConverters.package.enrichAsJavaIntFunction")
```

SethTisue:
I don't understand this. It seems to me we haven't touched this code.

I'm disinclined to spend my time digging into it. We're still in 0.x releases, and over at #213 we're planning on going 1.0, which means resetting binary compatibility anyway.

I'll just add an exclusion, and if it turns out this was a real problem we shouldn't have ignored, then I apologize to you, problem-experiencer-from-the-future.

PR #215, opened 2021-03-23
add Scala 3 to crossbuild

### LINKED ISSUE #200: Add Scala 3 crossbuild
volunteer?

it probably works with `withDottyCompat`, but it's better to have a "real" Scala 3 artifact

SethTisue:
fixes #200

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ import: scala/scala-dev:travis/default.yml
 language: scala
 
 scala:
+  - 3.0.0-RC1
   - 2.11.12
   - 2.12.13
   - 2.13.5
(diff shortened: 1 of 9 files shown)


SethTisue:
I did the build part, but now the actual code needs changes.

SethTisue:
This was routine work, with one exception: 5160b411d4076697478e2f0aa1239ecd9994ce8e, which removes a type parameter which I couldn't figure out how to make Scala 3 accept ("Polymorphic refinement method valueStepper without matching type in parent trait MapOps is no longer allowed")

...but I also don't really understand under what circumstances it might matter and I'm not inclined to really dig into it

@Ichoran Do you perhaps remember what that extra `V1 >: V` was doing there?

SethTisue:
I'm going to go ahead and merge this so we can roll 1.0.0-RC1 for Scala 3.0.0-RC2.

Still happy to receive feedback, post-merge.

